PR #37, opened 2017-10-18
Needed hypothesis and pytest to be installed to run test

fangohr:
using discretisedfield.test()

Suggest that we better move those into setup.py and remove
from test environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,11 +5,6 @@ RUN apt install -y git python3-pip curl
 RUN python3 -m pip install --upgrade pip pytest-cov nbval \
       matplotlib pyvtk git+git://github.com/joommf/joommfutil.git
 
-WORKDIR /usr/local
-#RUN git clone https://github.com/joommf/discretisedfield.git
-#RUN mkdir -p /usr/local/discretisedfield
-# guessing here cloned repository might be ...
-COPY discretisedfield/ /usr/local/discretisedfield/
-COPY * /usr/local/discretisedfield/
+COPY . /usr/local/discretisedfield/
 WORKDIR /usr/local/discretisedfield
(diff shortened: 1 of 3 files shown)


marijanbeg:
I wrote some hypothesis tests in discretisedfield (in a separate file test_mesh_new.py) only because I wanted to see how it would look like to have these types of tests for joommf. Hypothesis looks useful only in cases when we have simple input data, but in most of our functions we need to generate much more complex objects and this is not something that can be done with hypothesis elegantly.

In issue https://github.com/joommf/joommf/issues/28 we discussed before whether testing packages should be the dependencies of joommf packages. Our opinion then was that it should not - the same as in numpy and scipy. Any thoughts?

fangohr:
Regarding the question if we should introduce dependencies for tests to be run:

- for tests that are executable from within the module, I would say we must include the requirements. This is in inline with the numpy and script tests:

<pre>
In [1]: import numpy

In [2]: numpy.test()
Running unit tests for numpy
NumPy version 1.11.1
NumPy relaxed strides checking option: False
NumPy is installed in /Users/fangohr/anaconda3/lib/python3.5/site-packages/numpy
Python version 3.5.2 |Anaconda custom (x86_64)| (default, Jul  2 2016, 17:52:12) [GCC 4.2.1 Compatible Apple LLVM 4.2 (clang-425.0.28)]
nose version 1.3.7
.............................................................................................................................................................................................................................S...............................................
</pre>

- for tests that relate to the documentation I find this less obvious, and could be convinced either way. 
  If the dependency is easy to fulfil (say 'just another PyPI package') we might as well allow it.

fangohr:
Have added all the functionality from above via other merge requests. Will close this without merging.

PR #53, opened 2018-12-10
Ovf2vtk & k3d notebook

sergii-mamedov:


marijanbeg:
The realistic data from the `.omf` file can be read using `field = df.read('filename.omf')`.

discretisedfield/field.py
```diff
@@ -444,3 +447,48 @@ def _writeovf(self, filename, representation="txt"):
...
+        cmap = pylab.cm.get_cmap(name, n)
+        for i in range(cmap.N):
+            rgb = cmap(i)[:3]
+            data.append(rgb)
+
+        return np.array(data)
+
+
+    def plot3d(self, vector_scale=1.0, direction=0, color=0xff, colormap_name='', **kwargs):
+
```
fangohr: We need a documentation string, please.

discretisedfield/field.py
```diff
@@ -444,3 +447,48 @@ def _writeovf(self, filename, representation="txt"):
...
+        plot = k3d.plot()
+
+        # value of vectors
+        vectors = self.array.reshape(-1,3)
```
fangohr: PEP8 wants a space after a comma. Please run 'pep8' or 'flake' on this to make the source PEP8 compatible.

docs/dev/ovf2vtk.py
```diff
@@ -0,0 +1,42 @@
+import argparse
```
fangohr: Can we have at least one test for ovf2vtk? For example, conversion of an example file from ovf to vtk, and check the vtk file is created, and is likely to have the right data (for example check number of vectors in there). The very least this would detect if the script doesn't run.

discretisedfield/field.py
marijanbeg: It seems like this function was accidentally committed to this branch (and this pull request). This function was drafted as an exercise when @sergii-mamedov and @marijanbeg were working together. Its implementation and interface are part of Task 3 from the backlog and documentation is Task 5. In this pull request, I would like to stay concentrated on Tasks 2 and 6.

discretisedfield/field.py
marijanbeg: I agree. Running:
```
pycodestyle ovf2vtk.py
```
checks the PEP8. (`pep8` still exists, but `pycodestyle` is a new name for the package: https://github.com/PyCQA/pycodestyle/issues/466)

marijanbeg:
As I understand, the `ovf2vtk` tool is used as `python ovf2vtk.py ....`. This would mean that the user should clone the repository, in order to have the `ovf2vtk.py` file exposed. @fangohr, is this the functionality you had in mind? Can we make this to be a package/subpackage so that the call is `python -m ovf2vtk ...` or `python -m discretisedfield.ovf2vtk ...`. This way, it is enough only to install the package and it is not necessary to have the source.

PR #69, opened 2019-08-28
embed widget state in docs examples

minrk:
Using `Widgets > Save Notebook Widget State` menu embeds widget state in the doc, to allow widgets without a kernel

This is disabled by default because it guarantees a lot of version control churn, but is required if you want interactive widgets in static notebook rendering, e.g. readthedocs


marijanbeg:
Thank you @minrk!

PR #71, opened 2019-11-13
Plotting scanned line review

fangohr:
Make extraction of mx, my, mz more obvious

(based on the twilight_shift to plasma change; should just work if the other MR is accepted. 

Otherwise I'll fix it.)


marijanbeg:
Thank you @fangohr.

PR #75, opened 2020-08-20
Spin angles

lang-m:


marijanbeg:
Thank you, @lang-m. I had a look at the code and the things I would suggest doing are:

1. Make the behaviour of `norm` and `__abs__` the same. This way, we avoid possible confusion since the norm of a scalar field can be interpreted as an absolute value. I already changed that in this pull request and updated the documentation. After we merge, we are going to need to update the tutorial.

2. Topological charge methods are a bit messy at the moment because we have so many different methods, which complicates the code. My suggestion would be to keep only `topological_charge_density` and `topological_charge` methods in `Field` and add all the necessary functionality in those two functions. If necessary, not to make the methods too long, we can move some code to `utils`.

3. `spin_angles` and `max_spin_angles` do the job. However, I would probably try to simplify their connection to `angle`. Similarly, we can implement `max` and `min` methods for scalar fields, so that the `max_spin_angle` can be called as `field.angle(...).max`. Still not clear how this would work, but we can brainstorm that today.

marijanbeg:
I would also suggest ignoring the test for the moment until we agree on the right interface.

marijanbeg:
Can we base Berg-Luescher topological charge calculation on the density field? This way, we would avoid some repetition.

PR #79, opened 2021-01-22
Mm tools

lang-m:
Function names could be improved.

Missing:
- [ ] Tests

Tutorials
- [x] Topological charge
- [x] Cell angles
- [ ] Emergent magnetic field


lang-m:
We may think about also adding an option for plotting the whole region to k3d_subplots, similar to 
commit 7527f6666c7e9fb2026fbda83891a7ae85596618 . For k3d the region could be either semitransparent or shown as wireframe.

Issue #81, opened 2021-06-10
FFT


marijanbeg:
Implement FFT required by exsim.

marijanbeg:
This is being addressed in PR #83. As discussed, we need to extend FFT calculations for demag computations. Demag tensor can go to `df.tools`.

PR #83, opened 2021-06-14
WIP: exsim extensions.

lang-m:
Draft function definitions required for Fourier transform and complex fields as required for ubermag/exsim#5.


lang-m:
@marijanbeg Two things to think about:
- handling of complex field values (`dfu.as_array`)
- tolerance during mesh creation

marijanbeg:
Closes #81.

discretisedfield/plotting/mpl.py
```diff
@@ -0,0 +1,873 @@
...
+        self.planeaxis = dfu.raxesdict[data.mesh.attributes['planeaxis']]
+
+
+    # TODO: I assume the functionality from plot is being moved to plot_simplified?
+    def plot_simplified(self,
+                        ax=None,
+                        figsize=None,
+                        multiplier=None,
+                        scalar_args={},
+                        vector_args={},
```
lang-m, line 36: `scalar_args` and `vector_args` must be `None` because of some Python specialities. The default arguments are only evaluated once and reused for later calls. If e.g. `scalar_args` is modified during the first execution these changes will persist for later calls. Therefore, default arguments must generally be immutable (no `list`, `dict`, sometimes `tuple`, ...) unless this very special functionality is desired.

discretisedfield/field.py
```diff
@@ -3181,957 +3180,173 @@ def _fromhdf5(cls, filename):
...
+        if not self.mesh.attributes['isplane']:
+            msg = ('The field must be sliced before the Fourier transform can'
+                   'be computed in two dimensions.')
             raise ValueError(msg)
-
-        if self.dim != 1:
-            msg = f'Cannot plot dim={self.dim} field.'
+        
+        # TODO: Is it ever possible to have n.count(1) > 1 if filed is not sliced?
+        if self.mesh.n.count(1) > 1:
```
lang-m, line 3206: The problem here are fields that have been sliced multiple times, e.g. field.plane('x').plane('y').

discretisedfield/plotting/mpl.py
```diff
@@ -0,0 +1,873 @@
...
+        # We never allow plotting fields that are not sliced.
+        if not data.mesh.attributes['isplane']:
+            msg = 'The field must be sliced before it can be plotted.'
+            raise ValueError(msg)
+
+        self.data = data  # TODO: Consider renaming data to field.
+        self.planeaxis = dfu.raxesdict[data.mesh.attributes['planeaxis']]
+
+
+    # TODO: I assume the functionality from plot is being moved to plot_simplified?
```
lang-m, line 30: `plot_simplified` should completely replace `plot` (and then be renamed to `plot`).

discretisedfield/plotting/mpl.py
```diff
@@ -0,0 +1,873 @@
...
+                      if multiplier is None else None)
+
+        # Define defaults and update with user-passed dictionaries.
+        scalar_kwargs = {}
+        vector_kwargs = {'use_color': False, 'colorbar': False}
+        scalar_kwargs.update(scalar_args)
+        vector_kwargs.update(vector_args)
+
+        # Set up default scalar and vector fields.
+        # TODO: Do we allow user to specify what scalar and vector fields are? Did we have that before?
```
lang-m, line 119: We did previously allow user to specify `scalar_field` and `vector_field`. I have removed this functionality on purpose. The code is simpler without this functionality. In my opinion, `scalar` and `vector` should be used directly (together with a custom `ax`) if either `scalar_field` or `vector_field` should be different from the the default values of the `field` that is used when calling `field.mpl.plot_simplified()`.

discretisedfield/plotting/mpl.py
```diff
@@ -0,0 +1,873 @@
...
+        This is a convenience method used for quick plotting, and it combines
+        ``discretisedfield.plotting.Mpl.scalar`` and
+        ``discretisedfield.plotting.Mpl.vector`` methods. Depending on the
+        dimensionality of the field's value, it automatically determines what
+        plot is going to be shown. For a scalar field, only
+        ``discretisedfield.plotting.Mpl.scalar`` is used, whereas for a vector
+        field, both ``discretisedfield.plotting.Mpl.scalar`` and
+        ``discretisedfield.plotting.Mpl.vector`` plots are shown so that vector
+        plot visualises the in-plane components of the vector and scalar plot
+        encodes the out-of-plane component.
```
lang-m, line 49: Only vector is used for 2d fields.

discretisedfield/plotting/mpl.py
```diff
@@ -0,0 +1,873 @@
...
+        vector_kwargs.update(vector_args)
+
+        # Set up default scalar and vector fields.
+        # TODO: Do we allow user to specify what scalar and vector fields are? Did we have that before?
+        if self.data.dim == 1:
+            scalar_field = self.data
+            vector_field = None
+
+        elif self.data.dim == 2:
+            scalar_field = self.data
```
lang-m, line 125: 2D fields are not anymore supported by `scalar`. For 2d fields only `vector` should be used.

discretisedfield/plotting/mpl.py
```diff
@@ -0,0 +1,873 @@
...
+        scalar_field.mpl.scalar(ax=ax, multiplier=multiplier, **scalar_kwargs)
+        if vector_field is not None:
+            vector_field.mpl.vector(ax=ax, multiplier=multiplier, **vector_args)
+
+        self._axis_labels(ax, multiplier)
+
+        if filename is not None:
+            plt.savefig(filename, bbox_inches='tight', pad_inches=0.02)
+
+    def plot(self,
```
lang-m, line 149: Should be removed once we decide what exactly to support in `plot_simplified`. (Code is for now only kept for reference.)

discretisedfield/plotting/mpl.py
lang-m: Fixed in cad1126

discretisedfield/field.py
marijanbeg: Thanks. In that case, do we also consider that field to be sliced?

discretisedfield/plotting/mpl.py
marijanbeg: I completely agree!

discretisedfield/field.py
lang-m: Checks for slicing is in line 3200, right above.

PR #89, opened 2021-09-03
Vector component labels

lang-m:


discretisedfield/field.py
```diff
@@ -108,12 +109,24 @@ class Field:
...
+        if components is None:
+            if 2 <= dim <= 3:
+                components = ['x', 'y', 'z'][:dim]
+            elif dim > 3:
+                warnings.warn(f'Component labels must be specified for {dim=}'
+                              ' fields to get access to individual vector'
+                              ' components.')
+        if components is not None:
+            self.components = components
+
```
marijanbeg, line 129: This could be part of the property. No need to move it now, we can do that when we refactor.

discretisedfield/field.py
```diff
@@ -1383,7 +1460,8 @@ def __matmul__(self, other):
                        f'and {other.dim=} fields.')
                 raise ValueError(msg)
         elif isinstance(other, (tuple, list, np.ndarray)):
-            return self @ self.__class__(self.mesh, dim=3, value=other)
+            return self @ self.__class__(self.mesh, dim=3, value=other,
+                                         components=self.components)
```
marijanbeg, line 1464: This could think of making a partial function for avoiding this repetition for so many different methods. No need to do it now, we can do that when we refactor.

marijanbeg (approved):

PR #114, opened 2021-11-23
Simplify array initialisation

lang-m:
Breaks `as_array` function into multiple smaller ones to simplify and make it more readable.

- [x] add `dtype` to all `Field(...)` and `Field.__class__(...)` calls
- [x] update tests

### DIFF
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -1,3 +1,4 @@
+import collections
 import numbers
 import struct
 import warnings
@@ -15,7 +16,7 @@ import ubermagutil.typesystem as ts
 
 @ts.typesystem(mesh=ts.Typed(expected_type=df.Mesh, const=True),
                dim=ts.Scalar(expected_type=int, positive=True, const=True))
-class Field:
+class Field(collections.abc.Callable):  # could be avoided by using type hints
     """Finite-difference field.
 
     This class specifies a finite-difference field and defines operations for
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -435,7 +435,7 @@ class Field(collections.abc.Callable):  # could be avoided by using type hints
                 raise ValueError(msg)
 
             self.array /= self.norm.array  # normalise to 1
-            self.array *= dfu.as_array(val=val, self.mesh, dim=1,
+            self.array *= dfu.as_array(val, self.mesh, dim=1,
                                        dtype=self.dtype)
 
     def __abs__(self):
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -347,7 +347,7 @@ class Field(collections.abc.Callable):  # could be avoided by using type hints
 
     @array.setter
     def array(self, val):
-        self._array = dfu.as_array(self.mesh, self.dim, val)
+        self._array = dfu.as_array(self.mesh, self.dim, val, dtype=self.dtype)
 
     @property
     def norm(self):
@@ -435,7 +435,8 @@ class Field(collections.abc.Callable):  # could be avoided by using type hints
                 raise ValueError(msg)
 
             self.array /= self.norm.array  # normalise to 1
-            self.array *= dfu.as_array(self.mesh, dim=1, val=val)
+            self.array *= dfu.as_array(self.mesh, dim=1, val=val,
+                                       dtype=self.dtype)
 
     def __abs__(self):
         """Field norm.
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -109,11 +109,13 @@ class Field:
 
     """
 
-    def __init__(self, mesh, dim, value=0, norm=None, components=None):
+    def __init__(self, mesh, dim, value=0, norm=None, components=None,
+                 dtype=np.float64):
         self.mesh = mesh
         self.dim = dim
         self.value = value
         self.norm = norm
+        self.dtype = dtype
 
         self._components = None  # required in here for correct initialisation
         self.components = components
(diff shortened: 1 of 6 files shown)


discretisedfield/field.py
```diff
@@ -109,15 +109,19 @@ class Field:
...
-    def __init__(self, mesh, dim, value=0, norm=None, components=None):
+    def __init__(self, mesh, dim, value=0, norm=None, components=None,
+                 dtype=np.float64):
         self.mesh = mesh
         self.dim = dim
-        self.value = value
-        self.norm = norm
+        self.dtype = dtype
 
+        # order in here is important to avoid infinite recursion
```
marijanbeg, line 118: I am not sure I understand what is the recursion problem here. We could probably rewrite component setter.

discretisedfield/util/util.py
```diff
@@ -17,71 +18,89 @@ def array2tuple(array):
...
-        array = val
-    elif isinstance(val, (tuple, list, np.ndarray)):
-        if len(val) != dim:
-            msg = (f'Wrong dimension {len(val)} provided for value;'
-                   f' expected dimension is {dim}')
-            raise ValueError(msg)
-        if (isinstance(val, np.ndarray) and val.dtype == 'complex128'
-                or np.iscomplex(val).any()):
-            array = np.empty((*mesh.n, dim), dtype='complex128')
+def as_array(mesh, dim, val, dtype=np.float64):
```
marijanbeg, line 21: I would suggest removing `np.float64` as a default argument. This would force us to be explicit everywhere in the code and avoid confusions.

discretisedfield/util/util.py
```diff
@@ -17,71 +18,89 @@ def array2tuple(array):
...
+    array = np.empty((*mesh.n, dim), dtype=dtype)
```
marijanbeg, line 22: `np.full` (https://numpy.org/doc/stable/reference/generated/numpy.full.html) creates and fills an array at the same time.

discretisedfield/util/util.py
```diff
@@ -17,71 +18,89 @@ def array2tuple(array):
...
+
+
+@functools.singledispatch
+def _fill_array(val, array, mesh, dim):
+    raise ValueError('Unsupported type {type(val)}.')
+
+
+@_fill_array.register(numbers.Complex)  # contains numbers.Real
+def _(val, array, mesh, dim):
+    # TODO why not allow uniform values for vector fields
```
marijanbeg, line 33: We found that very confusing to pass `value=2` for a vector field.

discretisedfield/util/util.py
```diff
@@ -17,71 +18,89 @@ def array2tuple(array):
...
+    if dim > 1 and val != 0:
+        raise ValueError('Wrong dimension 1 provided for value;'
+                         f' expected dimension is {dim}')
+    array.fill(val)
+    return array
+
+
+@_fill_array.register(list)
+@_fill_array.register(tuple)
```
marijanbeg, line 42: Would `@_fill_array.register(collections.abc.Iterable)` work here?

discretisedfield/util/util.py
```diff
@@ -17,71 +18,89 @@ def array2tuple(array):
...
+def _(val, array, mesh, dim):
+    if len(val) != dim:
+        raise ValueError(f'Wrong dimension {len(val)} provided for value;'
+                         f' expected dimension is {dim}')
+    array[..., :] = val
```
marijanbeg, line 47: I guess that using `np.full` would be broadcasted correctly and this would probably reduce the number of separate implementations.

marijanbeg (commented):
Great idea! Added some (minor) comments.

lang-m:
@marijanbeg For simplicity I would suggest to discuss your comments in person in one of our next meetings.
@fangohr It might be good if you could have a quick look beforehand.

discretisedfield/util/util.py
lang-m: Currently array creation and filling are separated (on purpose):
- Creation (i.e. the shape) only depends on the mesh and datatype.
- Filling has multiple options but is always operating on the same array (shape and dtype).
- Different input for `val` (scalar vs array-like) needs different types of error handling - and I'd like to avoid `if`-`else` in favour of registered functions.
- `np.full` will work for certain `val` (e.g. scalar, array-like) but not for `callable` and `dict`
- To avoid code duplication in creating the array I've therefore split it up. Furthermore, we can simply add new implementations for filling and know that we get an array of the correct shape.

If we want to use `np.full` we should at least remove `_fill_array` and directly overload `as_array`.

marijanbeg: I think it would be clearer if we directly overloaded `as_array`.

fangohr:
The split of the lengthy function into multiple (single-dispatch) functions is a great improvement in reducing (perceived) complexity.

discretisedfield/util/util.py
lang-m: Allowing this would help in using `np.full` with as few checks as possible and relying on numpy to detect inconsistent shapes.

discretisedfield/util/util.py
```diff
@@ -17,71 +18,64 @@ def array2tuple(array):
...
+    raise ValueError('Unsupported type {type(val)}.')
+
+
+@as_array.register(numbers.Complex)
+@as_array.register(collections.abc.Iterable)
+def _(val, mesh, dim, dtype):
+    # we should think about allowing this as well
+    if isinstance(val, numbers.Complex) and dim > 1 and val != 0:
+        raise ValueError('Wrong dimension 1 provided for value;'
+                         f' expected dimension is {dim}')
```
lang-m, line 32: [Repeated from above] We should think about allowing scalar values for vector fields.

discretisedfield/field.py
lang-m: The recursion problem disappeared upon rewriting as_array

discretisedfield/util/util.py
fangohr: I can see why allowing this may be confusing. (Somebody may initialise a field with '3e10' - thanking this will act in the x-direction, because they compute everything with respect to an applied field in x. If the actual field value is [3e10, 3e10, 3e10], this is probably not what the user would expect. )

PR #118, opened 2022-01-14
Automatically determine dtype for array_like inputs.

lang-m:
Simplifies array creation and mathematical operations with fields.

Determining the array for `callable` or `dict` for `value` is more complex and therefore not done automatically. Instead, the numpy default (np.float64) is used by default. Furthermore, these types of value only occur during creation by the user where a different dtype can be easily specified (so we don't have any issues with wrong data types during operations on fields).

### DIFF
--- a/discretisedfield/util/util.py
+++ b/discretisedfield/util/util.py
@@ -30,11 +30,15 @@ def _(val, mesh, dim, dtype):
     if isinstance(val, numbers.Complex) and dim > 1 and val != 0:
         raise ValueError('Wrong dimension 1 provided for value;'
                          f' expected dimension is {dim}')
+    if dtype is None:
+        dtype = np.array(val).dtype
     return np.full((*mesh.n, dim), val, dtype=dtype)
 
 
 @as_array.register(collections.abc.Callable)
 def _(val, mesh, dim, dtype):
+    # will only be called on user input
+    # dtype must be specified by the user for complex values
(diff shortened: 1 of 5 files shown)


lang-m:
@marijanbeg @fangohr 
With the current implementation we can get
```python
>>> mesh = ...
>>> field = df.Field(mesh, dim=3, value=[1, 0, 0])
>>> type(field.array.dtype)
dtype('int64')
```
Maybe we should set a "lower boundary" for the type, i.e. use `float64` or more generic (complex) if required?

lang-m:
@marijanbeg Have you got any idea why the tests fail on Windows?

lang-m:
The most important changes are the one in `util.py`

discretisedfield/util/util.py
```diff
@@ -23,18 +23,27 @@ def as_array(val, mesh, dim, dtype):
...
+
 @as_array.register(numbers.Complex)
 @as_array.register(collections.abc.Iterable)
 def _(val, mesh, dim, dtype):
-    # we should think about allowing this as well
     if isinstance(val, numbers.Complex) and dim > 1 and val != 0:
         raise ValueError('Wrong dimension 1 provided for value;'
                          f' expected dimension is {dim}')
+    if dtype is None:
+        dtype = max(np.asarray(val).dtype, np.float64)
```
marijanbeg, line 39: I didn't know this was possible.

marijanbeg (approved):
Looks good to me.

PR #121, opened 2022-01-29
Rewrite ovf reading routine

lang-m:
Execution times for 1M cells:

- Reading
```
mode   old    new   speedup
==== ======= ====== =======
bin4 1730 ms  21 ms   82
bin8 1860 ms  36 ms   52
text 4920 ms 401 ms   12
```

- Writing
```
mode    old    new   speedup
==== ======== ====== =======
bin4 63000 ms  56 ms   1125
bin8 64000 ms  84 ms    762
text 69000 ms 4510 ms    15
```

Filesizes are
- `2.9M` for `bin4`
- `5.8M` for `bin8`
- `15M` for `txt`


discretisedfield/field.py
```diff
@@ -3059,112 +3061,98 @@ def _fromovf(cls, filename):
...
+                footer_length_estimate = 45
+                end = data.find(b'# End: Data Binary', footer_length_estimate)
+
+                data = data[:end].strip()  # trailing newline character in ovf
+                bdata = decoder.iter_unpack(data)
+                dtype = np.float64 if nbytes == 8 else np.float32
+                array = np.fromiter((e[0] for e in bdata), dtype=dtype)
+                array.shape = (-1, int(header['valuedim']))
+            else:
+                nrows = header['xnodes'] * header['ynodes'] * header['znodes']
```
marijanbeg, line 3139: Can we use len(mesh)?

discretisedfield/field.py
```diff
@@ -3059,112 +3061,98 @@ def _fromovf(cls, filename):
...
+                data = pd.read_csv(
+                    f,
+                    sep=' ',
+                    header=None,
+                    dtype=np.float64,
+                    skipinitialspace=True,
+                    nrows=nrows,
+                    comment='#')
+                array = np.stack([data[i].to_numpy() for i in data.columns],
+                                 axis=-1)
```
marijanbeg, line 3149: Can this be data.to_array().reshape(...)?

marijanbeg: This is to condense several lines to make it faster (if possible).

discretisedfield/field.py
```diff
@@ -3059,112 +3061,98 @@ def _fromovf(cls, filename):
...
-                    # metadata line', whereas mumax3 does not. Therefore if the
-                    # length of the data stream is not a multiple of nbytes, we
-                    # have to subtract the length of newline character from
-                    # data_end.
-                    if (data_end - data_start) % nbytes != 0:
-                        data_end -= len(nl)
+        mdata_str = ['meshunit', 'valueunit']  # information not used
+        mdata_list = ['valuelabels', 'valueunits']  # information not used
+
+        header = {}
```
marijanbeg, line 3072: I do not know if we can remove `mdata_float`, `mdata_int`, etc. and just use `header = {}`. This way we add all metadata information to `header`. Later, we can use `header.get('xmin', None)` when we need it. This might simplify the code and reduce line count.

discretisedfield/field.py
```diff
@@ -3059,112 +3061,98 @@ def _fromovf(cls, filename):
...
+        with open(filename, 'rb') as f:
+            # >>> READ HEADER <<<
+            header['ovf_version'] = next(f).decode('utf-8').split()[-1][-3]
```
marijanbeg, line 3075: ```header['ovf_version'] = int(next(f).decode('utf-8').split()[-1])``` or maybe ```*_, header['ovf_version'] = next(f).decode('utf-8').split()[-1]```. A lot of hardcoded indices are a bit difficult to read maybe.

discretisedfield/field.py
```diff
@@ -3059,112 +3061,98 @@ def _fromovf(cls, filename):
...
+            for line in f:
+                line = line.decode('utf-8')
+                if line.startswith('# Begin: Data'):
+                    mode = line.split()[3]
+                    if mode == 'Binary':
+                        nbytes = int(line.split()[4])
```
marijanbeg, line 3081: `nbytes = int(line.split()[-1])`

discretisedfield/field.py
marijanbeg: Also, we read all metadata to a dictionary and we can reuse it in the future more easily in the future.

discretisedfield/field.py
```diff
@@ -3059,112 +3061,98 @@ def _fromovf(cls, filename):
...
-        # extracted here.
-        if 'valuedim' not in mdatadict.keys():
-            mdatadict['valuedim'] = len(datalines) / len(mesh)
-
-        r_tuple = (*tuple(reversed(mesh.n)), int(mdatadict['valuedim']))
-        t_tuple = (*tuple(reversed(range(3))), 3)
-
-        return cls(mesh, dim=int(mdatadict['valuedim']),
-                   value=datalines.reshape(r_tuple).transpose(t_tuple))
+                information = line.split()
```
marijanbeg, line 3083: Can we `line.split(':')`?

discretisedfield/field.py
```diff
@@ -3059,112 +3061,98 @@ def _fromovf(cls, filename):
...
+                header['valuedim'] = 3
+
+            if mode == 'Binary':
+                if header['ovf_version'] == '1':
+                    endian = '>'  # ovf1 uses big-endian
+                else:
+                    endian = '<'  # ovf2 uses little-endian
+
+            # >>> MESH <<<
+            p1 = (header[key] for key in ['xmin', 'ymin', 'zmin'])
```
marijanbeg, line 3106: If we do not specify lists in the beginning, we can do type conversions at this point.

discretisedfield/field.py
```diff
@@ -3059,112 +3061,98 @@ def _fromovf(cls, filename):
...
+
+                # length of b'# End: Data Binary X\n# End: Segement\n' (37)
+                # plus additional safety
+                footer_length_estimate = 45
+                end = data.find(b'# End: Data Binary', footer_length_estimate)
+
+                data = data[:end].strip()  # trailing newline character in ovf
+                bdata = decoder.iter_unpack(data)
+                dtype = np.float64 if nbytes == 8 else np.float32
+                array = np.fromiter((e[0] for e in bdata), dtype=dtype)
```
marijanbeg, line 3136: very nice :)

marijanbeg: I do not know if this could help: https://numpy.org/doc/stable/reference/generated/numpy.fromfile.html

marijanbeg (commented):
Beautiful :) A few suggestions for you to consider. Please ignore if they do not make sense. Before merging, it would be cool if we could measure the execution time before and after using pandas/numpy ;)

discretisedfield/field.py
lang-m: By using the node information from the ovf file in here we have an additional safety check for the dimension (shape).

discretisedfield/field.py
lang-m: There is no `data.to_array()` method for a `pandas dataframe`. What exactly did you mean?

discretisedfield/field.py
lang-m: The different lists are mainly used to do the correct conversion (`float`, `int`, ...) right away. Somewhere we need to store the information of what to do with which value and with the current implementation all of this is in one place.

discretisedfield/field.py
lang-m: The file contains either `... v1.0\n"` or `"... 2.0\n"` without the `v` (!). I decide to only keep `"1"` or `"2"` (as `string`). Therefore all the hard-coded indices. So your suggestions both don't work. Maybe you have a different idea?

discretisedfield/field.py
lang-m: Switching to `np.fromfile` for binary data gave an additional speedup of ~8.

discretisedfield/field.py
```diff
@@ -3059,112 +3061,90 @@ def _fromovf(cls, filename):
...
+                test_data = decoder.unpack(f.read(nbytes))[0]
+                check = {4: 1234567.0, 8: 123456789012345.0}
+                if nbytes not in (4, 8) or test_data != check[nbytes]:
+                    raise ValueError(  # pragma: no cover
+                        f'Cannot read file {filename}. The file seems to be in'
+                        ' binary format ({nbytes} bytes) but the check value'
+                        f' is not correct: Expected {check[nbytes]}, got'
+                        f' {check}.')
+
+                array = np.fromfile(f, count=int(nodes * header['valuedim']),
```
marijanbeg, line 3128: Beautiful :)

discretisedfield/field.py
lang-m: I've moved it now.

discretisedfield/field.py
lang-m: @marijanbeg Is it better now?

lang-m:
Here is an example comparison for the execution time for 1e6 cells

```
mode   old    new   speedup
==== ======= ====== =======
bin4 1730 ms  21 ms   82
bin8 1860 ms  36 ms   52
text 4920 ms 401 ms   12
```

marijanbeg:
Wow! That is a serious speed-up. We should now definitely move to bin8 as a default.

lang-m:
[commit 9584e61 is not showing up]

lang-m:
@marijanbeg Speedup for the rewritten writing method (1M cells):
```
1 million cells 

mode    old    new   speedup
==== ======== ====== =======
bin4 63000 ms  56 ms   1125
bin8 64000 ms  84 ms    762
text 69000 ms 4510 ms    15
```

discretisedfield/field.py
```diff
@@ -2771,71 +2769,61 @@ def _writeovf(self, filename, representation='txt', extend_scalar=False):
...
+                f.write(struct.pack(bin_rep[representation][1],
+                                    bin_rep[representation][0]))
+
+                if extend_scalar:
+                    # remove scalar vector dimension
+                    reordered = reordered.reshape(list(reversed(self.mesh.n)))
+                    reordered = np.stack((reordered, np.zeros_like(reordered),
+                                          np.zeros_like(reordered)), axis=-1)
+
+                f.write(np.asarray(reordered,
```
marijanbeg, line 2813: I am wondering if https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tofile.html could be used here.

discretisedfield/field.py
```diff
@@ -2735,14 +2737,10 @@ def _writeovf(self, filename, representation='txt', extend_scalar=False):
...
-            raise TypeError(msg)
-
-        if extend_scalar and self.dim == 1:
-            write_dim = 3
-        else:
-            write_dim = self.dim
+        write_dim = 3 if extend_scalar and self.dim == 1 else self.dim
+        valuelabels = (' '.join(f'field_{c}' for c in self.components)
+                       if self.components else 'field_x')
+        valueunits = ' '.join(['None'] * self.dim)
```
marijanbeg, line 2743: I would probably move `if representation == ...` here to determine `repr_string`. This would allow building header list at once (less lines of code later).

discretisedfield/field.py
```diff
@@ -2771,71 +2769,61 @@ def _writeovf(self, filename, representation='txt', extend_scalar=False):
...
 
-        # Write data.
-        if representation in binary_reps:
-            # Reopen with binary write, appending to the end of the file.
-            with open(filename, 'ab') as f:
+        bin_rep = {'bin4': (1234567.0, '<f'),
+                   'bin8': (123456789012345.0, '<d')}
+
+        with open(filename, 'wb') as f:
+            f.write(bheader)
```
marijanbeg, line 2800: `f.write(''.join(f'# {line}\n' for line in header).encode('utf-8'))`

marijanbeg: One variable less

discretisedfield/field.py
```diff
@@ -2771,71 +2769,61 @@ def _writeovf(self, filename, representation='txt', extend_scalar=False):
...
+
+            if representation in bin_rep:
                 # Add the binary checksum.
-                packarray = [binary_reps[representation][0]]
+                f.write(struct.pack(bin_rep[representation][1],
+                                    bin_rep[representation][0]))
```
marijanbeg, line 2805: f.write(struct.pack(*bin_rep[representation]))

marijanbeg (commented):
Amazing improvement and simplification.

marijanbeg:
It is embarrassing to see that the writing routine took 60+ seconds to write the file. :(

discretisedfield/field.py
lang-m: From the documentation I've had the impression that it is not possible to control the byte order and precession. I'll have a closer look, maybe I misunderstood something.

discretisedfield/field.py
lang-m: I've changed it differently. What do you think?

marijanbeg: Even better :)

discretisedfield/field.py
lang-m: `ndarray.tofile` seems to be ~20% slower. So I decide to keep the current implementation.

marijanbeg:
Let's add a comment above that line that `ndarray.tofile` is about 20% slower to know in the future.

marijanbeg:
I am impressed by all the changes, please feel free to merge when you are ready ;)

fangohr:
Excellent. I am pleased the innocent suggestion of testing pandas to read text-based data has led to such speed-ups, also for binary data, and also better code readability.

I would suggest that we record the speed-ups you have measured in some kind of `changelog.txt` for this package:
- it will be useful for us to be able to look up at which version the speed improvements come in
- we should also mention the speed up in the next release (and then it will be good to be able to look up significant changes).

fangohr:
More than a remark than review feedback:

As we are drifting into performance behaviour of the source, we could introduce performance regression checks:
- have a set of tests that carry out reading and writing of data (as was done here)
- measure performance and record the execution time as part of the CI

This would catch deviations where, by accident, we have reduced execution performance. However, this is only useful if the hardware on which the tests run doesn't change. It increases additional complexity as we would need to track those execution times. On balance, I would say it is not worth doing this here.

lang-m:
@fangohr Changes for all packages are collected in the website repository, in particular the latest changes are in this PR: https://github.com/ubermag/ubermag.github.io/pull/2
The speedup is mentioned here: https://github.com/ubermag/ubermag.github.io/blob/b1e0532c4e3e48a5e15ffae2db76ebc62b8e8b67/source/changelog.rst (-> `discretisedfield` -> bullet point 4)
I can add more details about the actual numbers in there (i.e. show a table containing the numbers).

To reduce the number of files and keep it more clean I'd like to avoid introducing an additional file in `discretisedfield` (that we need to remember to look into before the next release).

fangohr:
> @fangohr Changes for all packages are collected in the website repository, in particular the latest changes are in this PR: [ubermag/ubermag.github.io#2](https://github.com/ubermag/ubermag.github.io/pull/2) The speedup is mentioned here: https://github.com/ubermag/ubermag.github.io/blob/b1e0532c4e3e48a5e15ffae2db76ebc62b8e8b67/source/changelog.rst (-> `discretisedfield` -> bullet point 4) I can add more details about the actual numbers in there (i.e. show a table containing the numbers).

Yes, please.
 
> To reduce the number of files and keep it more clean I'd like to avoid introducing an additional file in `discretisedfield` (that we need to remember to look into before the next release).

I agree that's a better solution.

lang-m:
Changelog is now updated.

PR #123, opened 2022-02-02
Add to_xarray as field method

swapneelap:
Adding `to_xarray` method for `Field` object which returns field value as an `xarray.DataArray`.
1. Typically, the returned DataArray has four dimensions, namely `x`, `y`, `z`, and `comp`. The first three corresponds to geometry while the fourth dimension corresponds to the components of field. `comp` dimension has `field.components` as co-ordinates.
2. If the `Field` object is a scalar field, the `comp` dimension is 'squeezed'  and the DataArray has only three dimensions corresponding to the geometry.
3. Instead of giving fourth dimension (i.e. `comp`) the name of the field, the name is assigned to the DataArray itself. The default name is `'field'`, but it can be changed with `name` parameter.
4. The units of the field can be set with `units` parameter. Units of geometry dimensions are set to `mesh.attributes['unit']` if the attribute exists, otherwise set to `'m'` (meter).


discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            raise TypeError(msg)
+
+        geo_dim = ['x', 'y', 'z']
+
+        data_array_coords = {dim: np.fromiter(self.mesh.axis_points(dim),
+                                              dtype=float)
+                             for dim in geo_dim
+                             }
+
+        geo_units_dict = {dim: 'm' for dim in geo_dim}  # fix units for geo
```
lang-m, line 3570: Use `mesh.attributes['unit']`.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+        if self.dim != 1:
+            data_array_dims = geo_dim + ['comp']
+            data_array_coords['comp'] = self.components
```
lang-m, line 3574: `self.components` could be `None` (for `self.dim > 3` where users have to set the names explicitly [but are not forced to do so during initialisation because this could break some of the operators])

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            field_array = self.array
+        else:
+            data_array_dims = geo_dim
+            field_array = np.squeeze(self.array, self.array.ndim - 1)
```
lang-m, line 3578: `np.squeeze(self.array, axis=-1)`

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+        data_array = xr.DataArray(field_array,
+                                  dims=data_array_dims,
+                                  coords=data_array_coords)
+
+        for dim in geo_units_dict:
+            data_array[dim].attrs['units'] = geo_units_dict[dim]
```
lang-m, line 3585: `unit` without `s`?

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        ``name`` and ``units``. If the type of value passed to any of the two
+        arguments is not string, then a TypeError is raised.
+
+        Parameters
+        ----------
+        name : str, optional
+
+            String to set name of the field ``DataArray``
+
+        units : str, optional
```
lang-m, line 3499: `unit` without `s`? We cannot have multiple different units.

lang-m (commented):
Looks good. I left a few minor comments.

lang-m:
Please all ignore the failing `pre-commit.ci`. This isn't fully set up yet and therefore completely meaningless.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         else:
             return self.__class__(mesh[0], dim=result.shape[-1], value=result,
                                   components=self.components)
+
+    def to_xarray(self, name='field', units=None):
+        """Field value as ``xarray.DataArray``.
+
+        The function returns a ``xarray.DataArray`` with dimensions ``x``,
+        ``y``, ``z``, and ``comp`` (only if vector field). The coordinate of
```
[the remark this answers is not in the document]
swapneelap: Yes, done.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        the geometric dimensions are derived from ``self.mesh.axis_points``,
+        and for vector field components from ``self.components``. The ``units``
+        attribute of geometric dimensions is set to ``'m'``.
```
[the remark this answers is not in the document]
swapneelap: This is idea hit me at the time of writing the PR description :). I have changed the default value to `elf.mesh.attributes['unit']`

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        ...
+        TypeError: name argument must be a string.
+        >>> field.to_xarray(units={'field': 'A/m'})
+        Traceback (most recent call last):
+        ...
+        TypeError: units argument must be a string.
+
+        """
+
+        if type(name) != str:
```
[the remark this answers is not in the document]
swapneelap: Changed.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            msg = "name argument must be a string."
+            raise TypeError(msg)
+        elif units is not None and type(units) != str:
```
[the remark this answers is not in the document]
swapneelap: Changed it as well. I have also changed `elif` to `if` as the variables in question are different.

discretisedfield/field.py
swapneelap: Done.

discretisedfield/field.py
swapneelap: Thank you mentioning this. I was not aware of it. I was under the impression that for `self.dim>3` it was commendatory to pass components list. This could have potentially broke `to_xarray` as one cannot set `None` as a dimension co-ordinate. I have changed the code accordingly.

discretisedfield/field.py
swapneelap: Thanks, I have changed it.

discretisedfield/field.py
swapneelap: The `units` attribute comes from netCDF convention that xarray follows as mentioned [here](https://xarray.pydata.org/en/stable/getting-started-guide/quick-overview.html#attributes). xarray uses this attribute mostly as plot labels.

discretisedfield/field.py
lang-m: It is possible to create higher-dimensional fields implicitly and therefore `components` cannot be mandatory:
```python
>>> f1 = df.Field(..., dim=2, ...)  # has components x, y [automatically]
>>> f2 = df.Field(..., dim=2, ...)  # has components x, y [automatically]
>>> f1 << f2            # stack fields; we can neither pass component-labels manually
Field(..., dim=4, ...)  # nor use the existing ones as x, y are would appear twice.

```

discretisedfield/field.py
lang-m: I wasn't aware of this standard. We should keep `units` everywhere.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        """
+        if not isinstance(name, str):
+            msg = "name argument must be a string."
+            raise TypeError(msg)
+
+        if units is not None and not isinstance(units, str):
+            msg = "units argument must be a string."
+            raise TypeError(msg)
+
+        geo_dim = ['x', 'y', 'z']
```
marijanbeg, line 3562: Maybe renaming `geo_dim` to `axes` for consistency with the rest of the code?

marijanbeg (commented):
Great code @swapneelap. I added a few very minor comments.

discretisedfield/field.py
swapneelap: Good point! I have changed the variable name to `axes`.

discretisedfield/field.py
swapneelap: I understand, thank you for the example.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        ...
+        TypeError: name argument must be a string.
+        >>> field.to_xarray(units={'field': 'A/m'})
+        Traceback (most recent call last):
+        ...
+        TypeError: units argument must be a string.
+
+        """
+        if not isinstance(name, str):
+            msg = "name argument must be a string."
```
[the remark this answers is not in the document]
swapneelap: Changed.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            raise TypeError(msg)
+
+        if units is not None and not isinstance(units, str):
+            msg = "units argument must be a string."
```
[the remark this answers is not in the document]
swapneelap: Changed

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+            Field values DataArray.
+
+        Raises
+        ------
+        TypeError
+
+            If either ``name`` or ``units`` argument is not a string.
+
+        Examples
```
fangohr, line 3516: Excellent we have examples.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        Field(...)
+        >>> field.to_xarray()
+        <xarray.DataArray 'field' (x: 10, y: 10, z: 10, comp: 3)>
+        ...
+        Coordinates:
+          * x        (x) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+          * y        (y) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+          * z        (z) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+          * comp     (comp) <U1 'x' 'y' 'z'
+        >>> field.to_xarray().sel(comp='x')
```
fangohr, line 3536: Minor point: How about writing 

```python
xa = field.to_xarray()
xa.sel(comp='x')
```
instead of 
```python
field.to_xarray().sel(comp='x')
```
?

No change in meaning, but as an example for inexperienced users, we would want to culture the idea of creating the xarray object once, then working with it.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,117 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        <xarray.DataArray 'field' (x: 10, y: 10, z: 10)>
+        ...
+        Coordinates:
+          * x        (x) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+          * y        (y) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+          * z        (z) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+            comp     <U1 'x'
+        >>> field.to_xarray(name=21)
```
fangohr, line 3544: I would omit the demonstration of the exceptions being raised: Due to the multiple ways of selecting data from xarrays, I worry that somebody may misunderstand the example as another way of selecting data, which for some reason doesn't work.

I don't think we need to demo the failure. (We should have  a test that checks the exception type and that it is raised instead.)

pyproject.toml
```diff
@@ -16,7 +16,8 @@ authors = [
     {name = "Martin Lang"},
     {name = "Samuel Holt"},
     {name = "Ryan A. Pepper"},
-    {name = "Hans Fangohr"}
+    {name = "Hans Fangohr"},
+    {name = "Swapneel Amit Pathak"}
```
fangohr, line 20: Does the order matter? If so, let's move line 20 to be between 17 and 18.

fangohr:
This looks good. 

Do we have tests for this method yet? If not, we need one. Martin or I can help to suggest some.

swapneelap:
> This looks good.
> 
> Do we have tests for this method yet? If not, we need one. Martin or I can help to suggest some.

No, I have not added the tests yet. Yes, it's a good idea to discuss them!

discretisedfield/field.py
swapneelap: Good point! I have changed the docstring accordingly.

discretisedfield/field.py
swapneelap: Changed.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,115 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+        axes = ['x', 'y', 'z']
+
+        data_array_coords = {
+            axis: np.fromiter(self.mesh.axis_points(axis), dtype=float)
+            for axis in axes
+        }
+
+        geo_units_dict = None
+        if 'unit' in self.mesh.attributes:
```
swapneelap, line 3563: @lang-m I found that it is possible to create mesh without `unit` key in `mesh.attributes`; example code is below:
```python
p2=(10,10,10)
p1=(0,0,0)
cell = (1,1,1)
region = df.Region(p1, p2)
mesh = df.Mesh(region=region, cell=cell, attributes={'units': 'meter'})
```
Thus `geo_units_dict = {axis: self.mesh.attributes['unit'] for axis in axes}` throws error.

swapneelap:
I have added the tests for `to_xarray` function as we discussed in the meeting. Let me know what you think.

marijanbeg:
Thanks a lot @swapneelap. What about `field.xarray` instead of `field.to_xarray`? We do not have `to_` in other methods.

lang-m:
I think I would prefer `to_xarray` for two reasons:
- This method is different from others e.g. `field.array` where we only pass the underlying `numpy` array. Here we create a different "independent" object. @swapneelap Something to test (or maybe you know): Does `xarray` create a copy of the data during creation, i.e. if we (1) create a field, (2) create the xarray, (3) modify the old field (e.g. `field.array = ...`) does this affect the xarray object?
- We now have `discretisedfield -> xarray`. When we have a second method for `xarray -> discretisedfield` I think it would be most consistent if the methods are called `field.to_xarray()` and `Field.from_xarray(<DataArray>)` (a classmethod).

discretisedfield/field.py
lang-m: Maybe we can use `m` (metre) as default if there is not `mesh.attributes['unit']`?

lang-m: Something like
```python
geo_units_dict = {
    axis: self.mesh.attributes['unit']
          if 'unit' in self.mesh.attributes
          else 'm'
    for axis in axes
}
```

lang-m:
I've added a reduced version of our new pre-commit configuration to avoid that the tests always show up as failing.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,115 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            geo_units_dict = {
+                axis: self.mesh.attributes['unit'] for axis in axes
+            }
+
+        if self.dim != 1:
```
lang-m, line 3568: I think `self.dim > 1` is a bit faster to understand.

discretisedfield/field.py
```diff
@@ -3475,3 +3476,115 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            data_array_dims = axes
+            field_array = np.squeeze(self.array, axis=-1)
+
+        data_array = xr.DataArray(field_array,
+                                  dims=data_array_dims,
+                                  coords=data_array_coords)
+
+        if geo_units_dict is not None:
+            for dim in geo_units_dict:
+                data_array[dim].attrs['units'] = geo_units_dict[dim]
```
lang-m, line 3583: I think we should always have some default units (see above).

discretisedfield/tests/test_field.py
```diff
@@ -2201,3 +2200,78 @@ def test_numpy_ufunc(self):
...
+                )
+                assert np.array_equal(
+                    np.array(list(f.mesh.axis_points('z'))),
+                    fxa.z.values
+                )
+                assert np.array_equal(f.array, fxa.values)
+                assert all(
+                    fxa[i].attrs['units'] == f.mesh.attributes['unit']
+                    for i in ['x', 'y', 'z']
+                )
```
lang-m, line 2227: Shorter:
```python
for i in 'xyz':
    assert np.array_equal(
        np.array(list(f.mesh.axis_points(i))),
        fxa[i].values
    )
    assert fxa[i]['units'] == f.mesh.attributes['unit']
assert np.array_equal(f.array, fxa.values)
```

discretisedfield/tests/test_field.py
```diff
@@ -2201,3 +2200,78 @@ def test_numpy_ufunc(self):
...
+                    fxa.x.values
+                )
+                assert np.array_equal(
+                    np.array(list(f.mesh.axis_points('y'))),
+                    fxa.y.values
+                )
+                assert np.array_equal(
+                    np.array(list(f.mesh.axis_points('z'))),
+                    fxa.z.values
+                )
```
lang-m, line 2245: See above

discretisedfield/tests/test_field.py
```diff
@@ -2201,3 +2200,78 @@ def test_numpy_ufunc(self):
...
+                assert 'comp' not in fxa.dims
+                assert np.array_equal(f.array.squeeze(axis=-1), fxa.values)
+                assert all(
+                    fxa[i].attrs['units'] == f.mesh.attributes['unit']
+                    for i in ['x', 'y', 'z']
+                )
```
lang-m, line 2251: Merge with loop above (see above)

discretisedfield/tests/test_field.py
```diff
@@ -2201,3 +2200,78 @@ def test_numpy_ufunc(self):
...
+
+        f6d = self.pf << self.pf
+        f6d_xa = f6d.to_xarray()
+        assert f6d_xa['comp'].size == 6
```
lang-m, line 2255: Can we test the default (i.e. not set) `comp` labels?

lang-m (commented):
The tests look already quite good. I left a few comments, mainly to shorten it (without removing anything).

discretisedfield/field.py
```diff
@@ -3475,3 +3476,115 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+          * y        (y) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+          * z        (z) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+            comp     <U1 'x'
+
+        """
+        if not isinstance(name, str):
+            msg = "Name argument must be a string."
+            raise TypeError(msg)
+
+        if units is not None and not isinstance(units, str):
```
marijanbeg, line 3551: Is just `if not isinstance(units, str)` a sufficient condition?

discretisedfield/field.py
```diff
@@ -3475,3 +3476,115 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        axes = ['x', 'y', 'z']
+
+        data_array_coords = {
+            axis: np.fromiter(self.mesh.axis_points(axis), dtype=float)
+            for axis in axes
+        }
+
+        geo_units_dict = None
+        if 'unit' in self.mesh.attributes:
+            geo_units_dict = {
```
marijanbeg, line 3564: `geo_units_dict = dict.fromkeys(axes, self.mesh.attributes['unit'])`

discretisedfield/field.py
```diff
@@ -3475,3 +3476,115 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        if geo_units_dict is not None:
+            for dim in geo_units_dict:
+                data_array[dim].attrs['units'] = geo_units_dict[dim]
+
+        if units is not None:
+            data_array.attrs['units'] = units
+
+        data_array.name = name
+
+        return data_array
```
marijanbeg, line 3590: It might be (structurally) cleaner if we create DataArray at this point and pass all necessary arguments at once (such as `attr`, `name`).

discretisedfield/tests/test_field.py
```diff
@@ -2201,3 +2200,78 @@ def test_numpy_ufunc(self):
...
+                    fxa[i].attrs['units'] == f.mesh.attributes['unit']
+                    for i in ['x', 'y', 'z']
+                )
+                assert all(fxa['comp'].values == f.components)
+
+            for value, dtype in self.sfuncs:
+                f = df.Field(mesh, dim=1, value=value, dtype=dtype)
+                fxa = f.to_xarray()
+                assert isinstance(fxa, xr.DataArray)
+                assert np.array_equal(
```
marijanbeg, line 2234: Is maybe `np.allclose` safer to use here?

swapneelap:
@lang-m 
>  This method is different from others e.g. `field.array` where we only pass the underlying `numpy` array. Here we create a different "independent" object. @swapneelap Something to test (or maybe you know): Does `xarray` create a copy of the data during creation, i.e. if we (1) create a field, (2) create the xarray, (3) modify the old field (e.g. `field.array = ...`) does this affect the xarray object?

If possible, the view is used instead of a copy when initializing `xarray.DataArray` as mentioned [here](https://xarray.pydata.org/en/stable/generated/xarray.DataArray.html). I also checked it with `id()` and I get the same object for `field.array` and `DataArray.values`.

swapneelap:
> Thanks a lot @swapneelap. What about `field.xarray` instead of `field.to_xarray`? We do not have `to_` in other methods.

@marijanbeg I was thinking along the lines of what @lang-m has pointed out. Even though `field.array` and `DataArray.values` are the same object [whenever possible](https://xarray.pydata.org/en/stable/generated/xarray.DataArray.html), the fact that output of `to_xarray` can hold metadata and has numerous methods of its own somehow distances itself from other field object methods in my mind. Moreover, the name is convenient if we plan to add `from_xarray` class method.

discretisedfield/field.py
swapneelap: Good point! I have changed this.

discretisedfield/field.py
swapneelap: I have removed the `data_array.name = name`, and instead, passed it directly like `xr.DataArray(..., name=name)`. Also, I have removed the condition that checks if `units` argument is `None`. I pass it anyway like `xr.DataArray(..., attrs=dict(units=units))`. I think having `units` attribute as None is harmless. 
```Python
for dim in geo_units_dict:
            data_array[dim].attrs['units'] = geo_units_dict[dim]
```
However, I am not able to figure out how to set the `units` attribute for axes before initializing the DataArray. Let me know if you have any ideas.

discretisedfield/field.py
swapneelap: The default value of `units` is `None`. `if not isinstance(units, str)` will raise `TypeError` for `units=None`

discretisedfield/field.py
swapneelap: Good point! Changed.

discretisedfield/field.py
swapneelap: Changed

discretisedfield/tests/test_field.py
swapneelap: Good point, thank you! I have changed the code accordingly.

discretisedfield/tests/test_field.py
swapneelap: Done

discretisedfield/tests/test_field.py
swapneelap: Done

discretisedfield/tests/test_field.py
swapneelap: Added 
```Python 
assert 'comp' not in f6d_xa.coords
```

discretisedfield/tests/test_field.py
swapneelap: Although it is not giving me any error with `array_equal` but I understand your point. Should I replace all `array_equal` with `allclose`?

discretisedfield/field.py
lang-m: Something like `isinstance(None, str)` [-> `False`] should work just fine.

discretisedfield/field.py
lang-m: I think the only option would be to first create `DataArrays` for `coords` (where we could pass `attrs`). But I wouldn't suggest doing this as it will complicate the code rather than simplifying it.

discretisedfield/tests/test_field.py
lang-m: @marijanbeg We use the same generator to get `fxa.x.values` and all numpy arrays use `np.float64` (i.e. the same precision as the generated Python `float`s). Can't we then expect that both arrays are **exactly** the same (i.e. the same bytes)?

discretisedfield/field.py
```diff
@@ -3475,3 +3476,115 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+          * comp     (comp) <U1 'x' 'y' 'z'
+        >>> xa = field.to_xarray()
+        >>> xa.sel(comp='x')
+        <xarray.DataArray 'field' (x: 10, y: 10, z: 10)>
+        ...
+        Coordinates:
+          * x        (x) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+          * y        (y) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+          * z        (z) float64 0.5 1.5 2.5 3.5 4.5 5.5 6.5 7.5 8.5 9.5
+            comp     <U1 'x'
```
lang-m, line 3544: These lines are quite `fragile` during doctests (i.e. they tend to break tests easily and potentially could do so if something changes in the representation of xarray in the future. Maybe we should either show less details or ignore some of the output during testing. Maybe something like
```python
>>> xa = field.to_xarray()
>>> xa  # doctest: +SKIP
<xarray.DataArray 'fi..
...
Coordinates:
[rest of the output]
>>> xa.sel([one specific point])
[1, 0, 0]  # that might be the wrong format
>>> xa.shape
[shape]
>>> xa.sel(comp='x').shape
[shape]
```

discretisedfield/field.py
```diff
@@ -3475,3 +3476,115 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        ...
+        >>> p1 = (0, 0, 0)
+        >>> p2 = (10, 10, 10)
+        >>> cell = (1, 1, 1)
+        >>> mesh = df.Mesh(p1=p1, p2=p2, cell=cell)
+        >>> field = df.Field(mesh=mesh, dim=3, value=(1, 0, 0), norm=1.)
+        ...
+        >>> field
+        Field(...)
+        >>> field.to_xarray()
```
lang-m, line 3528: We could have additional "headings" (the snippet isn't properly highlighted on GitHub, headings are `rst`):

```python
1. [Create a field]
>>> import discretisedfield as df
[...]
2. [Create `xarray.DataArray`]
>>> xa = field.to_xarray()
>>> xa
[...]
3. Extract a single point
>>> xa.sel(...)
[...]
[and so on]
```
(see also my comment below)

discretisedfield/field.py
swapneelap: The output `DataArray` from the updated `to_xarray` function always has `units` attribute which was absent in the docstring. I have fixed it and pushed the fix. This should get ride of the doctest errors. I prefer showing less than skipping the tests. Skipping the tests might lead to display of wrong information in docs. Or should I keep it as is since the error is fixed?

discretisedfield/field.py
swapneelap: Good idea! I have added headings.

discretisedfield/field.py
swapneelap: Exactly what I was thinking. It looks like there is no simple way of doing it.

discretisedfield/field.py
swapneelap:
```Python
if not isinstance(units, str):
            msg = "Units argument must be a string."
            raise TypeError(msg)
```
@lang-m  Sorry, I did not get what you mean. The code above raises `TypeError` if I try `field.to_xarray()` (without the `units` argument). Am I missing something obvious?

lang-m: @swapneelap The code *should raise* a `TypeError` if you do not pass `units` (i.e. the default `units = None`) because you manually raise it in the last line (the message would be important but it will for sure be what you manually set for `msg`).

lang-m: The actual issue is slightly different (I missed this in the beginning). We want `units` to be optional (but of type `str` if passed). So we need to keep the current implementation (only raising if not `None` and not `str`).

swapneelap:
@lang-m Whilst writing `from_xarray` class method, I realized that it is impossible to reconstruct the original mesh from the `DataArray` if any of the geometric axes has a single cell. Since, we just have the value of the *midpoint* of the cell (as the co-ordinate value in the `DataArray`) for those axes, we have no idea about the *actual size* of the cell.

I can think of two solutions:
- We store the value of `self.mesh.cell` as a `DataArray` attribute when using `to_xarray` function. Similarly, we ask the user to provide a `DataArray` which has the same `cell` attribute when using `from_xarray` class method.
-   We ask the user to provide the cell size as a `DataArray` attribute when using `from_xarray` only when one or more geometric axes have one cell.

PS: The tests will fail until we fix this. If you I can edit the tests to ignore this.

marijanbeg:
Hi @swapneelap, thank you for working on this. I guess we would need both to allow people to reconstruct the field object from xarray even if the xarray was not created with `to_xarray` method.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        return data_array
+
+    @classmethod
+    def from_xarray(cls, xa):
+        if not isinstance(xa, xr.DataArray):
+            msg = "Argument must be a xr.DataArray."
+            raise TypeError(msg)
+
+        if xa.ndim not in [3, 4]:
+            msg = "DataArray dimensions must be 3 or 4."
```
marijanbeg, line 3610: Is 3 for scalar and 4 for vector field? This error message might be confusing to users. (what is a 4D field)

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            raise ValueError(msg)
+
+        if xa.ndim == 3:
+            for dim in xa.dims:
+                if dim not in ['x', 'y', 'z']:
```
marijanbeg, line 3615: `if any(dim not in 'xyz' for dim in xa.dims):`

marijanbeg: Or maybe even `if xa.ndim == 3 and any(dim not in 'xyz' for dim in xa.dims):`

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                    msg = "The dimensions must be either 'x', 'y', or 'z'."
+                    raise ValueError(msg)
+        elif xa.ndim == 4:
+            for dim in xa.dims:
+                if dim not in ['x', 'y', 'z', 'comp']:
```
marijanbeg, line 3620: Similar to previous suggestion

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                    msg = "The dimensions must be either "\
+                          "'x', 'y', 'z' or 'comp'."
+                    raise ValueError(msg)
+            if xa['comp'].values.size != len(set(xa['comp'].values)):
+                msg = "All the components should be unique."
+                raise ValueError(msg)
+            if 'comp' in xa.coords:
+                for comp in xa['comp'].values:
+                    if not isinstance(comp, str):
```
marijanbeg, line 3629: `if 'comp' in xa.coords and any(not isinstance(comp, str) for comp in xa['comp'].values)`

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        if xa.values.dtype not in [np.float64, np.complex128]:
+            msg = "DataArray values must be either "\
+                  "np.float64 or np.complex128."
+            raise ValueError(msg)
+
+        cell = tuple()
+        p1 = tuple()
+        p2 = tuple()
+        for i in 'xyz':
+            dec = 5  # minimum decimal places to round to.
```
marijanbeg, line 3654: Does this work for nanometers?

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            return cls(mesh=mesh,
+                       dim=1,
+                       value=np.expand_dims(xa.values, axis=-1),
+                       dtype=xa.values.dtype)
+
+        else:
+            return cls(mesh=mesh,
+                       dim=xa['comp'].values.size,
+                       value=xa.values,
+                       dtype=xa.values.dtype)
```
lang-m, line 3689:
```python
comp = xa.comp.values if 'comp' in xa.coords else None
values = np.expand_dims(xa.values, axis=-1) if xa.ndim == 3 else xa.values
dim = 1 if xa.ndim == 3 else values.shape[-1]
return cls(mesh=mesh, dim=dim, value=values, dtype=xa.values.dtype)
```

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        cell = tuple()
+        p1 = tuple()
+        p2 = tuple()
+        for i in 'xyz':
+            dec = 5  # minimum decimal places to round to.
+
+            if xa[i].values.size > 1:
+                diff = np.diff(xa[i].values).mean()
+            else:
+                diff = np.float64(1e-9)  # imput value for cell dimension.
```
lang-m, line 3659: We should store [allways] (and expect) the cell size in the attributes if (at least) one direction has only a single cell.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+            exp = np.log10(np.abs(diff))
+            if exp < 0:
+                dec = dec + int(np.abs(exp).round())
+
+            diff = diff.round(decimals=dec)
+            cell = cell + (abs(diff),)
+            p1 = p1 + ((xa[i].values[0] - diff * 0.5).round(decimals=dec),)
+            p2 = p2 + ((xa[i].values[-1] + diff * 0.5).round(decimals=dec),)
```
lang-m, line 3668: Do we have to round?
```python
try:
    cell = np.array([xa[i].values[1] - xa[i].values[0] for i in 'xyz'])
except IndexError:
    cell = xa.attrs['cell']  # maybe wrong syntax (?)
p1 = np.array([xa[i].values[0] for i in 'xyz']) - cell / 2
p2 = np.array([xa[i].values[-1] for i in 'xyz']) + cell / 2
```

discretisedfield/tests/test_field.py
```diff
@@ -2201,3 +2200,204 @@ def test_numpy_ufunc(self):
...
+        args = [int(), float(), str(), list(), dict(), xr.Dataset(),
+                np.empty((20, 20, 20, 3))]
+        bad_value1 = xr.DataArray(np.ones((20, 20, 5, 3), dtype=np.int8),
+                                  dims=['x', 'y', 'z', 'comp'],
+                                  coords=dict(x=np.arange(0, 20),
+                                              y=np.arange(0, 20),
+                                              z=np.arange(0, 5),
+                                              comp=['x', 'y', 'z']),
+                                  name='mag',
+                                  attrs=dict(units='A/m'))
```
lang-m, line 2294: That should be valid in my opinion.

discretisedfield/tests/test_field.py
```diff
@@ -2201,3 +2200,204 @@ def test_numpy_ufunc(self):
...
+                                       attrs=dict(units='A/m'))
+
+        bad_coord_comp2 = xr.DataArray(np.ones((20, 20, 20, 3), dtype=float),
+                                       dims=['x', 'y', 'z', 'comp'],
+                                       coords=dict(x=np.arange(0, 20),
+                                                   y=np.arange(0, 20),
+                                                   z=np.arange(0, 20),
+                                                   comp=['a', 'b', 'c']),
+                                       name='mag',
+                                       attrs=dict(units='A/m'))
```
lang-m, line 2369: This refers to my other comment. Why would we not allow this?

lang-m (commented):
We should store the cell size in the attributes dictionary and expect it to present when (at least) one direction only has a single cell.

discretisedfield/field.py
swapneelap: Yes, the dimensions of the input `DataArray` are `x`, `y`, `z` (corresponding to the geometry), and `comp` (corresponding to the field). If we have the scalar field, than `comp` is not present and we have just the tree dimensions corresponding to the geometry. If we have a vector field (could be n dimensional), than `comp` is present which will hold the components of the field. Hence, I wanted to check that we have either three or four dimensions for scalar field or n dimensional vector field respectively.

I agree that the message is a bit opaque. Maybe we can output, "DataArray dimensions should be 3 for a scalar field and 4 for a vector field."?

discretisedfield/field.py
swapneelap: Good point! Changed

discretisedfield/field.py
swapneelap: Changed

discretisedfield/field.py
swapneelap: Good point! But I will not be able to use `elif` here which checks, even if `comp` are `str`, they should be one of `xyz`?

discretisedfield/field.py
swapneelap: Yes, it works for nanometers because of:
```Python
exp = np.log10(np.abs(diff))
if exp < 0:
    dec = dec + int(np.abs(exp).round())
```
Here, the code calculates the exponent of the `diff` (essentially the cell dimension) and adds it to `dec`. So, for nanometer dimensions, `dec = 5 + 9` which is 14.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        for dim in geo_units_dict:
+            data_array[dim].attrs['units'] = geo_units_dict[dim]
+
+        return data_array
+
+    @classmethod
+    def from_xarray(cls, xa):
+        if not isinstance(xa, xr.DataArray):
+            msg = "Argument must be a xr.DataArray."
+            raise TypeError(msg)
```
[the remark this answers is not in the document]
swapneelap: Good point! Changed

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+        if xa.ndim not in [3, 4]:
+            msg = "DataArray dimensions must be 3 or 4."
+            raise ValueError(msg)
+
+        if xa.ndim == 3:
+            for dim in xa.dims:
+                if dim not in ['x', 'y', 'z']:
+                    msg = "The dimensions must be either 'x', 'y', or 'z'."
```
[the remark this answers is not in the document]
swapneelap: Good point! Changed.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                    raise ValueError(msg)
+        elif xa.ndim == 4:
+            for dim in xa.dims:
+                if dim not in ['x', 'y', 'z', 'comp']:
+                    msg = "The dimensions must be either "\
+                          "'x', 'y', 'z' or 'comp'."
```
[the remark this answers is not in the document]
swapneelap: Changed.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                    raise ValueError(msg)
+            if xa['comp'].values.size != len(set(xa['comp'].values)):
+                msg = "All the components should be unique."
```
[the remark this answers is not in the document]
swapneelap: Changed.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                raise ValueError(msg)
+            if 'comp' in xa.coords:
+                for comp in xa['comp'].values:
+                    if not isinstance(comp, str):
+                        msg = "The values of 'comp' must be a string."
```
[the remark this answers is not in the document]
swapneelap: Changed! Looks like cancelling Grammarly subscription was a bad move :/.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                        raise ValueError(msg)
+                    elif comp not in ['x', 'y', 'z']:
```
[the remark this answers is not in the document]
swapneelap: changed

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                        msg = "The values of 'comp' must be either "\
+                              "'x', 'y', or 'z'."
+                        raise ValueError(msg)
+
+        for i in 'xyz':
+            if xa[i].values.size != 1:
+                if not np.allclose(np.diff(xa[i].values),
+                                   np.diff(xa[i].values).mean()):
+                    msg = f'Co-ordinates of {i} must be'\
+                          ' equally spaced.'
```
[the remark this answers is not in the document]
swapneelap: Good point! Changed it.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                    raise ValueError(msg)
+
+        if xa.values.dtype not in [np.float64, np.complex128]:
+            msg = "DataArray values must be either "\
+                  "np.float64 or np.complex128."
+            raise ValueError(msg)
```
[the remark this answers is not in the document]
swapneelap: I initially wanted to filter out `str` data type, but then I realized that `xr.DataArray` can hold many data types (all that `np.ndarray` can hold). Hence, as a first step, I kept the condition on data type to be very rigid. I was thinking we can discuss and make it more pragmatic.

Also, from the `test_from_xarray_invalid_args_and_DataArrays`, the `bad_value2.values` has `dtype('<U1')` (Which I think is `unicode`?). This passes the test:
```Python
if xa.values.dtype == str:
            msg = "DataArray values must not be strings."
            raise ValueError(msg)
```
Let me know if you have some ideas.

discretisedfield/field.py
swapneelap: Changed.

discretisedfield/tests/test_field.py
swapneelap: Yes, I agree. We should include integer data types.

discretisedfield/field.py
swapneelap: I initially did not round. But the values of `p1`, `p2`, and `cell` of original mesh and the mesh re-created by `from_xarray`do not match. I observed deviations like `0` becoming `1.02753e-23` and `5e-9` becoming `5.0000000000001e-9` after performing the floating point arithmetic.

These deviations give error when comparing the original field and the one re-created by `from_xarray` (even with `field.allclose()` method) as the undying mesh do not match. This error is mitigated by rounding but then the issue is to what decimal place to round.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,210 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                raise ValueError(msg)
+            if 'comp' in xa.coords:
+                for comp in xa['comp'].values:
+                    if not isinstance(comp, str):
+                        msg = "The values of 'comp' must be a string."
+                        raise ValueError(msg)
+                    elif comp not in ['x', 'y', 'z']:
+                        msg = "The values of 'comp' must be either "\
+                              "'x', 'y', or 'z'."
+                        raise ValueError(msg)
```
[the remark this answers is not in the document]
swapneelap: I understand that this is not a restriction when creating the field object. However, I was under the impression that `mpl` method uses `axes.quiver` for plotting the arrows which takes into account only the Cartesian co-ordinates. Hence, I wanted to put a restriction on the values of `comp`. However, there is mistake here that I notice. I should have checked `if xa['comp'].values.size in [2, 3]:`.

Going forward, either I can remove this condition (which I prefer) or restrict the `comp` values to `x`, `y`, and `z` if we are dealing with 2D or 3D field.

discretisedfield/field.py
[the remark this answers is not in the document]
lang-m: The `<U1` is `unsigned integer with 1 byte and littel-endian`, see here: https://numpy.org/doc/stable/reference/arrays.interface.html#python-side (`typestr`).

discretisedfield/field.py
lang-m: Could we use `dec = np.finfo(xa.values.dtype).precision`?

discretisedfield/field.py
```diff
@@ -3479,3 +3480,192 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+    @classmethod
+    def from_xarray(cls, xa):
+        if not isinstance(xa, xr.DataArray):
+            raise TypeError("Argument must be a xr.DataArray.")
+
+        if xa.ndim not in [3, 4]:
+            raise ValueError("DataArray dimensions must be 3 for a scalar "
+                             "and 4 for a vector field.")
+
+        if xa.ndim == 3 and any(dim not in 'xyz' for dim in xa.dims):
```
lang-m, line 3612: [I've already mentioned this but maybe it got lost] I think
```python
if xa.ndim == 3 and sorted(xa.dims) != ['x', 'y', 'z']:
```
would be even more readable and simpler. (Similar below for `xa.dim == 4`)

discretisedfield/field.py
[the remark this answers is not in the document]
lang-m: The plotting does not use the `component` labels.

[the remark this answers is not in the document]
lang-m: I would suggest just removing this check.

discretisedfield/field.py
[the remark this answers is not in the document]
lang-m: We can safely ignore the datatype at this point. The checks will automatically happen during creation of the `Field` object.

discretisedfield/field.py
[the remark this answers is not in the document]
swapneelap: Removed.

discretisedfield/field.py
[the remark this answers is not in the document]
swapneelap: Removed.

discretisedfield/field.py
swapneelap: Yes, this works perfectly! I have changed it.

discretisedfield/field.py
swapneelap: Sorry, I missed it. Changed.

discretisedfield/field.py
```diff
@@ -3603,49 +3603,40 @@ def to_xarray(self, name='field', units=None):
...
-            if 'comp' in xa.coords:
-                for comp in xa['comp'].values:
-                    if not isinstance(comp, str):
-                        msg = "The values of 'comp' must be a string."
-                        raise ValueError(msg)
-                    elif comp not in ['x', 'y', 'z']:
-                        msg = "The values of 'comp' must be either "\
-                              "'x', 'y', or 'z'."
-                        raise ValueError(msg)
+                raise ValueError("All the components must be unique.")
```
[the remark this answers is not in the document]
swapneelap: Removed.

discretisedfield/field.py
swapneelap: Added the functionality.

discretisedfield/field.py
swapneelap: Using `dec = np.finfo(xa[i].values.dtype).precision` now as suggested by @lang-m.

discretisedfield/field.py
```diff
@@ -3479,3 +3480,165 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            field_array = self.array
+        else:
+            data_array_dims = axes
+            field_array = np.squeeze(self.array, axis=-1)
+
+        data_array = xr.DataArray(field_array,
+                                  dims=data_array_dims,
+                                  coords=data_array_coords,
+                                  name=name,
+                                  attrs=dict(units=units, cell=self.mesh.cell))
```
lang-m, line 3582: We should include `cell` in the tests.

swapneelap: Added

discretisedfield/field.py
```diff
@@ -3479,3 +3480,165 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        if xa.ndim == 3 and sorted(xa.dims) != ['x', 'y', 'z']:
+            raise ValueError("The dimensions must be 'x', 'y', and 'z'.")
+        elif xa.ndim == 4 and sorted(xa.dims) != ['comp', 'x', 'y', 'z']:
+            raise ValueError("The dimensions must be 'x', 'y', 'z',"
+                             "and 'comp'.")
+
+        for i in 'xyz':
+            if xa[i].values.size > 1 and not np.allclose(
+                    np.diff(xa[i].values), np.diff(xa[i].values).mean()):
+                raise ValueError(f'Co-ordinates of {i} must be'
```
marijanbeg, line 3607: Co-ordinates -> Coordinates

swapneelap: Changed

discretisedfield/field.py
```diff
@@ -3479,3 +3480,258 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+        The class method accepts an ``xarray.DataArray`` as an argument to
+        return a ``discretisedfield.Field`` object. The DataArray must have
+        either three (``x``, ``y``, and ``z`` for a scalar field) or four
+        (additionally ``comp`` for a vector field) dimensions corresponding to
+        geometric axes and components of the field, respectively. The
+        coordinates of the ``x``, ``y`` and ``z`` dimensions represent the
+        discretisation along the respective axis and must have equally spaced
+        values. The coordinates of ``comp`` represent the field components
+        (for eg. ['x', 'y', 'z']).
```
lang-m, line 3606: either `for example` or `e.g.`

discretisedfield/field.py
```diff
@@ -3479,3 +3480,258 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        ):
+            try:
+                cell = xa.attrs['cell']
+            except KeyError:
+                raise KeyError(
+                    "DataArray must have a 'cell' attribute if any "
+                    "of the geometric directions has a single cell."
+                ) from None
+        else:
+            cell = [np.diff(xa[i].values).mean() for i in 'xyz']
```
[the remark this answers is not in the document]
swapneelap: Good idea! changed

discretisedfield/field.py
```diff
@@ -3479,3 +3480,258 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                                             p2=self.mesh.region.p2))
+
+        for dim in geo_units_dict:
+            data_array[dim].attrs['units'] = geo_units_dict[dim]
+
+        return data_array
+
+    @classmethod
+    def from_xarray(cls, xa):
+        """Create Field from ``xarray.DataArray``
```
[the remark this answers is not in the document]
swapneelap: Changed

discretisedfield/field.py
```diff
@@ -3479,3 +3480,258 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+        The DataArray is expected to have ``cell``, ``p1`` and ``p2``
+        attributes for creating ``discretisedfield.Mesh`` required by the
+        ``discretisedfield.Field`` object. However, in the absence of these
+        attributes, the coordinates of ``x``, ``y``, and ``z`` dimensions are
+        utilized. It should be noted that ``cell`` attribute is a must if any
+        of the geometric directions has only a single cell.
+
+        Failing to meet the above mentioned conditions will lead to a KeyError
+        in the absence of ``cell`` attribute or a ValueError for the rest.
+        TypeError is raised if input argument is not an ``xarray.DataArray``.
```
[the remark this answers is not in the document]
swapneelap: Removed

discretisedfield/field.py
```diff
@@ -3479,3 +3480,258 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+        Raises
+        ------
+        TypeError
+
+            If argument is not ``xarray.DataArray``.
+
+        KeyError
+
+            If at least one of the geometric dimension has only a single cell
```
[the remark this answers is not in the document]
swapneelap: Changed

discretisedfield/field.py
```diff
@@ -3479,3 +3480,258 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            and ``cell`` attribute is absent.
```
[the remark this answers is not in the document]
swapneelap: Changed

discretisedfield/field.py
```diff
@@ -3479,3 +3480,258 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+        2. Create Field from DataArray
+
+        >>> import discretisedfield as df
+        ...
+        >>> fld = df.Field.from_xarray(xa)
+        >>> fld
+        Field(...)
+        >>> fld.average
+        (1.0, 1.0, 1.0)
```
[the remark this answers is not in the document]
swapneelap: Good point! Changed

discretisedfield/field.py
swapneelap: Changed

discretisedfield/tests/test_field.py
```diff
@@ -2201,3 +2200,207 @@ def test_numpy_ufunc(self):
...
+                    )
+                    assert fxa[i].attrs['units'] == f.mesh.attributes['unit']
+                assert 'comp' not in fxa.dims
+                assert np.array_equal(f.array.squeeze(axis=-1), fxa.values)
+
+        f6d = self.pf << self.pf
+        f6d_xa = f6d.to_xarray()
+        assert f6d_xa['comp'].size == 6
+        assert 'comp' not in f6d_xa.coords
+
```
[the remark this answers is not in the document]
swapneelap: Good point! Added.

marijanbeg (approved):
This looks good to me - great work! We can further simplify this code when @lang-m finishes implementing midpoints and vertices :)

lang-m (commented):
The documentation looks good to me, it is very readable. I've removed the `print` for the xarray DataArrays to get the nicer-looking html representation and tagged the relevant cells to avoid failures of nbval.

I would suggest to keep this PR open until we can use the new `mesh.midpoints`.

swapneelap:
> The documentation looks good to me, it is very readable. I've removed the `print` for the xarray DataArrays to get the nicer-looking html representation and tagged the relevant cells to avoid failures of nbval.
> 
> I would suggest to keep this PR open until we can use the new `mesh.midpoints`.

Thank you, that sounds good. I will update the methods when you merge the changes.

lang-m (approved):

marijanbeg:
Very nice @swapneelap!

swapneelap:
Thank you all for your reviews and comments! It helped in making the methods concise, readable, and efficient. :)

PR #125, opened 2022-02-18
Update repository metadata

lang-m:


marijanbeg:
Do we need the two `*.txt` files to be committed?

lang-m:
No, that was a mistake. I'll remove them. Thanks for mentioning.

PR #129, opened 2022-03-11
New implementation to read and write VTK files

lang-m:


discretisedfield/field.py
```diff
@@ -3153,6 +3204,54 @@ def _fromvtk(cls, filename):
...
+            components = None
+
+        value = vns.vtk_to_numpy(array).reshape(*reversed(n), dim)
+        value = value.transpose((2, 1, 0, 3))
+
+        mesh = df.Mesh(p1=p1, p2=p2, n=n)
+        return cls(mesh, dim=dim, value=value, components=components)
+
+    @classmethod
+    def _fromvtk_legacy(cls, filename):
```
marijanbeg, line 3250: I would probably put this to `df.tools`.

discretisedfield/mesh.py
```diff
@@ -396,6 +396,13 @@ def axis_points(self, axis, /):
         for i in range(self.n[axis]):
             yield self.index2point((0, 0, 0))[axis] + i*self.cell[axis]
 
+    def cell_points(self, axis, /):
```
marijanbeg, line 399: What do you think about making this `mesh.x`, `mesh.y`... similar to `field.x`?

marijanbeg (commented):
Very good, I have added a few comments

discretisedfield/field.py
```diff
@@ -8,6 +8,8 @@
 import numpy as np
 import pandas as pd
 import ubermagutil.typesystem as ts
+import vtk
+from vtk.util import numpy_support as vns
```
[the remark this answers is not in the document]
lang-m: `numpy_support` still is a sub-module, not a function. And something like `vtk.util.numpy_support.numpy_to_vtk(...)` is (almost) too long for our short line length. Therefore the abbreviation to `vns.numpy_to_vtk(...)`. What do you think?

discretisedfield/field.py
```diff
@@ -2834,18 +2838,73 @@ def _writeovf(self, filename, representation='bin8', extend_scalar=False):
 
             f.write(bfooter)
 
-    def _writevtk(self, filename):
+    def _to_vtk(self):
```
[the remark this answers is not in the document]
lang-m: This function could be potentially useful in the future. As we can now easily convert a field to vtk we could think about adding support for interactive 3d plots based on e.g. pyvista ([demo](https://docs.pyvista.org/user-guide/jupyter/index.html)) in addition to `k3d`. That could also just be a notebook demonstrating how to do it. So instead of hiding it or moving to `util` would rather think about making it public right away.

discretisedfield/field.py
lang-m: Everything in `df.tools` currently is public - which this function should never be. It should always only be used automatically if we detect that we cannot read the file with the new function. If moving it at all I think `df.util` would be a better place.

discretisedfield/field.py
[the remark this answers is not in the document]
marijanbeg: Sure, sounds good. I was just worried giving users too many options to read/save the file.

discretisedfield/field.py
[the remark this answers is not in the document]
marijanbeg: Does `import vtk.util.numpy_support as vns` work?

discretisedfield/field.py
marijanbeg: I was thinking like... you tried to open a VTK file that was written a long time ago... you cannot and that functionality is now deprecated. However, you can find a "tool" in "df.tools" that helps you with that. I would really like to avoid cluttering Field class, especially because the chances of anyone needing this function is infinitesimally small.

discretisedfield/mesh.py
lang-m: A different interface for this would be good. Also because `cell_points` and `axis_points` is very similar.

An additional slight difficulty is the fact that we now add support for `xarray`. `xarray` also supports getting coordinates (similar to `axis_points`) of the centre points of cells, e.g. `<xarray-field>.x`. So it could be confusing if `<mesh>.x` returns different points (and a different shape `n+1`)

Maybe we could have something like the following instead.
- Alternative 1 (no `mesh.x`)
  ```python
  >>> mesh = df.Mesh(...)
  >>> mesh.midpoints.x  # replaces axis_points
  <generator ...>
  >>> mesh.vertices('x')  # replaces cell_points
  <generator ...>
  ```
  As indicated `x` could either be passed as a string to the two methods or we could use e.g. `namedtuple`s and have properties.
- Alternative 2
  ```python
  >>> mesh = df.Mesh(...)
  >>> mesh.x
  <edge object>  # new object
  >>> mesh.x.vertices  # replaces cell_points
  <generator ...> 
  >>> mesh.x.midpoints  # replaces axis_points
  <generator ...>
  ```

discretisedfield/field.py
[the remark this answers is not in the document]
lang-m: Shall we then make it public right away?

discretisedfield/field.py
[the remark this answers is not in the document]
lang-m: It does.

discretisedfield/field.py
lang-m: Then maybe we can just move it to `df.util`? That way we could still use it automatically (which is only two lines of code) (and the import is already there) without the need for the user to do anything manually. Explaining to users (or even a deprecation warning) would be more lines (of code) that just using the function from `df.util`.

discretisedfield/mesh.py
marijanbeg: I like `vertices` vs. `midpoints` to explain the difference. The interface would be then, e.g. `mesh.vertices.x`, `mesh.midpoints.z`,... Namedtuple should be the best way forward here.

discretisedfield/mesh.py
```diff
@@ -386,15 +389,51 @@ def axis_points(self, axis, /):
...
+        >>> list(mesh.vertices.x)
+        [0, 2, 4, 6, 8, 10]
+
+        """
+        vertices = collections.namedtuple('vertices', ['x', 'y', 'z'])
 
-        for i in range(self.n[axis]):
-            yield self.index2point((0, 0, 0))[axis] + i*self.cell[axis]
+        return vertices(*(np.linspace(pmin, pmax, n + 1) for pmin, pmax, n in
+                          zip(self.region.pmin, self.region.pmax, self.n)))
```
lang-m, line 436: @marijanbeg If you are happy with this (not anymore lazy) calculation of the `vertices`/`midpoints` we can merge the PR.

lang-m: [Copied from Zulip, looks like you've missed it there, and slightly extended]
I think using `np.linspace` would be a lot better, because:
- it is generally more convenient to use (no need to convert to a list first) and all parts within `discretisedfield` can use/need a numpy array "version" anyways.
- faster for larger n (example: 2000 cells, generator: 750μs, numpy: 30 μs (raw numpy), 200 μs with the additional calls to the mesh/region); subsequent calls are ~50 ns (due to caching).
- easier to use because we can assign and reuse it, whereas more care is necessary with the generator expression.

PR #134, opened 2022-03-23
Correct aspect ratio for mesh and region mpl calls.

lang-m:
There is a problem with overlapping ticks and axis labels. However, the benefit
of quickly seeing the shape from the plot (without reading axis ticks) should
outweight this issue. If required, one can always manually set better
ticks/labels after creating the plot.

Old behavior (can be obtained by explicitly passing the new parameter `box_aspect`, `(4, 4, 3)` is the `matplotlib` default):
grafik

New behaviour (obtained with `box_aspect='auto'`, which is the default):
grafik

### DIFF
--- a/discretisedfield/mesh.py
+++ b/discretisedfield/mesh.py
@@ -1207,7 +1207,7 @@ class Mesh:
         return df.Field(self, dim=3, value=dn, norm=norm)
 
     def mpl(self, *, ax=None, figsize=None, color=dfu.cp_hex[:2],
-            multiplier=None, filename=None, **kwargs):
+            multiplier=None, box_aspect='auto', filename=None, **kwargs):
         """``matplotlib`` plot.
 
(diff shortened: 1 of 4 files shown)


discretisedfield/mesh.py
```diff
@@ -1334,14 +1343,21 @@ def mpl_subregions(self, *, ax=None, figsize=None, color=dfu.cp_hex,
...
-
-            If filename is passed, the plot is saved. Defaults to ``None``.
-
         show_region : bool, optional
 
             If ``True`` also plot the whole region. Defaults to ``False``.
 
+        box_aspect : str, array_like (3), optional
+
+            Set the aspect-ratio of the plot. If set to `auto` the aspect ratio
```
marijanbeg, line 1352: ``'auto'``

marijanbeg:
Very nice!

PR #135, opened 2022-03-23
New initialisation method when passin a `df.Field` using the new xarray functionality.

lang-m:
Test
- field with `mesh.n = (100, 100, 10)`
- creation of a new field with `mesh.n = (10, 10, 10)` and passing the old field to `value`

Performance improvement:
- old implementation (using that the field is callable): ~ 6.5 s
- new implementation (interpolation done by `xarray`): ~ 5 ms

@swapneelap This is probably the first use case for your new method.

### DIFF
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -263,8 +263,8 @@ class Field:
         .. seealso:: :py:func:`~discretisedfield.Field.array`
 
         """
-        value_array = as_array(self._value, self.mesh, self.dim,
-                               dtype=self.dtype)
+        value_array = _as_array(self._value, self.mesh, self.dim,
+                                dtype=self.dtype)
         if np.array_equal(self.array, value_array):
             return self._value
         else:
@@ -273,7 +273,7 @@ class Field:
     @value.setter
     def value(self, val):
         self._value = val
-        self.array = as_array(val, self.mesh, self.dim, dtype=self.dtype)
+        self.array = _as_array(val, self.mesh, self.dim, dtype=self.dtype)
 
     @property
     def components(self):
@@ -362,7 +362,7 @@ class Field:
 
     @array.setter
     def array(self, val):
-        self._array = as_array(val, self.mesh, self.dim, dtype=self.dtype)
+        self._array = _as_array(val, self.mesh, self.dim, dtype=self.dtype)
 
     @property
     def norm(self):
@@ -450,7 +450,7 @@ class Field:
                 raise ValueError(msg)
 
             self.array /= self.norm.array  # normalise to 1
-            self.array *= as_array(val, self.mesh, dim=1, dtype=None)
+            self.array *= _as_array(val, self.mesh, dim=1, dtype=None)
 
     def __abs__(self):
         """Field norm.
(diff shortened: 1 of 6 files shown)


swapneelap:
@lang-m Happy to see that it is useful to the project! :)
Great performance improvement BTW.

discretisedfield/field.py
```diff
@@ -3794,3 +3795,88 @@ def from_xarray(cls, xa):
...
+
+
+@_as_array.register(collections.abc.Callable)
+def _(val, mesh, dim, dtype):
+    # will only be called on user input
+    # dtype must be specified by the user for complex values
+    array = np.empty((*mesh.n, dim), dtype=dtype)
+    for index, point in zip(mesh.indices, mesh):
+        res = val(point)
+        array[index] = res
```
marijanbeg, line 3829: `array[index] = val(point)`

discretisedfield/field.py
```diff
@@ -3794,3 +3795,88 @@ def from_xarray(cls, xa):
...
+@_as_array.register(Field)
+def _(val, mesh, dim, dtype):
+    if mesh.region not in val.mesh.region:
+        raise ValueError(f'The region {val.mesh.region} of the provided field'
+                         f' does not contain the region {mesh.region} of the'
+                         ' field that is being created.')
+    value = val.to_xarray().sel(x=mesh.midpoints.x,
+                                y=mesh.midpoints.y,
+                                z=mesh.midpoints.z,
+                                method='nearest').data
```
marijanbeg, line 3842: Beautiful!

discretisedfield/field.py
```diff
@@ -3794,3 +3795,88 @@ def from_xarray(cls, xa):
...
+        return value.reshape(mesh.n + (-1,))
+    return value
+
+
+@_as_array.register(dict)
+def _(val, mesh, dim, dtype):
+    # will only be called on user input
+    # dtype must be specified by the user for complex values
+    if dtype is None:
+        dtype = np.float64
```
marijanbeg, line 3854: `dtype = dtype or np.float64` (short-circuiting)

discretisedfield/field.py
```diff
@@ -3794,3 +3795,88 @@ def from_xarray(cls, xa):
...
+    if 'default' in val and not callable(val['default']):
+        fill_value = val['default']
+    else:
+        fill_value = np.nan
```
marijanbeg, line 3858: single-line?

marijanbeg (commented):
Looks very good! Just added a few minor suggestions.

Is there a reason why we want all `_as_array` in `Field` class and not in `dfu`?

lang-m:
@marijanbeg Could you please have a look at the last test failure. This is a problem with floating-point accuracy. My suggestion would be to slightly relax `region.__contains__` by adding `numpy.isclose` tests for `pmin` and `pmax`. This would mean going from `other.pmin >= self.pmin` to `other.pmin > or "approx equal" self.pmin`. Have you got any other ideas?

marijanbeg:
@lang-m, I agree. I am just not sure what is the right way of avoiding hard-coded tolerances. Maybe we could define some tolerances in `utils` and use them throughout `df`.

lang-m:
@marijanbeg Do you think sticking to the default settings for `rtol` and `atol` of `np.isclose` wouldn't be sufficient?

lang-m:
The new `_as_array` implementation depending on `df.Field` cannot be in `dfu` because we run into issues with circular imports. To make it consistent I decided to move everything to the `field` module as this also is the only place where we need theses functions.

marijanbeg:
Nice, I like that.

marijanbeg:
@lang-m, default value for `rtol` is fine. Only `atol` could be an issue because the values we have in the field can be in a wide range. Probably setting it to zero is fine.

lang-m:
@marijanbeg Setting `atol` to 0 does not work when one of the points is at `(0, 0, 0)` (or probably also just 0 in one direction, I didn't check that). I've now introduced a different way of calculating `rtol` and `atol`. Could you please review [d789e1c](https://github.com/ubermag/discretisedfield/pull/135/commits/d789e1c833006523bb32968b62e182915ddd834f).

marijanbeg:
It does exactly what it says on the tin.

PR #137, opened 2022-04-12
Test calculation of the demag field.

lang-m:
Tests are performed for a sphere and an approximation of an (infinite) 2d plane.
The results are compared against analytical results for the ideal cases in the
centre of the sphere/plane (which requires quite large tolerances) and against
pre-computed results obtained with OOMMF.


samjrholt (approved):
All looks good to me. It is good to see the tests on both the demag tensor and the demag field as compared to oommf.

marijanbeg:
Looks good to me - well spotted! Could we have at least 1-sentence docstrings for `_g`, `_f` just to be sure we know what those functions are doing?

PR #140, opened 2022-04-20
Add holoviews based plotting methods

lang-m:
- [x] Scalar
- [x] Vector
- [x] Vector with out-of-plane colouring
~- [ ] Lightness~
- [x] Contour


lang-m:
I would suggest to ignore the lightness plot for now and later add it in a separate PR.

discretisedfield/plotting/hvplot_field.py
```diff
@@ -0,0 +1,452 @@
...
+
+        All the default values can be changed by passing dictionaries to
+        ``scalar_kw`` and ``vector_kw``, which are then used in subplots.
+
+        Therefore, to understand the meaning of the keyword arguments which can be
+        passed to this method, please refer to
+        ``discretisedfield.plotting.HvplotField.scalar`` and
+        ``discretisedfield.plotting.HvplotField.vector`` documentation. Filtering of the
+        scalar component is applied by default (using the norm for vector fields,
+        absolute values for scalar fields). To turn of filtering add ``{'filter_field':
```
samjrholt, line 59: To turn off filtering ...

samjrholt (commented):
Looks good Martin. The only comment I have is if it is worth doing something similar to the axis SI prefixes as we do in mpl? As for small length scales we currently have it in scientific notation and (m) rather than e.g. (nm)

samjrholt:
Also, is there an interpolation method for hvplot?

samjrholt (approved):

PR #141, opened 2022-04-25
Field unit

lang-m:
We need to think about how to properly handle the units in mathematical operations.


discretisedfield/field.py
```diff
@@ -3230,10 +3259,39 @@ def _fromovf(cls, filename):
...
+        except KeyError:
+            units = None
+        else:
+            if len(unit_list) == 0:
+                units = None  # no unit in the file
+            elif len(set(unit_list)) != 1:
+                warnings.warn(
+                    f"File {filename} contains multiple units for the individual"
+                    f" components: {unit_list=}. This is not supported by"
+                    " discretisedfield. Units are set to None."
```
fangohr, line 3283: Nice.

discretisedfield/field.py
```diff
@@ -3230,10 +3259,39 @@ def _fromovf(cls, filename):
...
+                components = None
+
+        try:
+            unit_list = header["valueunits"].split()
+        except KeyError:
+            units = None
+        else:
+            if len(unit_list) == 0:
+                units = None  # no unit in the file
+            elif len(set(unit_list)) != 1:
```
swapneelap, line 3279: We are not checking the same if a user directly assigns multiple units to a field like `df.Field(mesh, dim=3, value=(1, 1, 1), units="m s kg")`. Do you think we should check for multiple units at the time of initialization as well?

swapneelap (commented):
The notebook looks fine. I have added a few comments/questions. Everything else looks good to me.

discretisedfield/field.py
lang-m: For the OOMMF generated files we know that we can get the same unit multiple (typically three) times separated by spaces and we only want to keep it one time. For a general field we could expect a situation in which a user wants to define a "more complicated" unit consisting of multiple SI units and decide to separate theses with spaces (instead of asterisks) - which should be fine as the units are basically only used for plots generated via `xarray`. That's why I didn't add any more strict tests for the user input. [I expect (and hope) that OOMMF would never write a single combined unit with a space in it.]

discretisedfield/field.py
```diff
@@ -3429,6 +3487,7 @@ def fftn(self):
             dim=len(values),
             value=np.stack(values, axis=3),
             components=self.components,
+            units=self.units,
```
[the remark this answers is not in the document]
lang-m: You're right, we shouldn't keep the units after an FT.

discretisedfield/field.py
swapneelap: I see. Sounds reasonable.

PR #145, opened 2022-05-21
Fix running multiple test commands via invoke

lang-m:


lang-m:
Already done on master.

PR #147, opened 2022-06-12
WIP: activate colour in pytest output in github actions

fangohr:
First test to see if this works (activate colour in pytest output in github actions).

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -5,6 +5,9 @@ on:
   schedule:
     - cron: '0 0 * * 1'  # run every Monday
 
+env:
+  PYTEST_ADDOPTS: "--color=yes"
+
 jobs:
   workflow:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


lang-m:
@fangohr I have updated the workflow file on `master`. Could you please try again.

lang-m:
On other pull requests they are now running twice (pull_request and push).

fangohr:
> On other pull requests they are now running twice (pull_request and push).

Hm, yes - I have seen that in many projects. One would think there is a simple solution to this (but I haven't investigated at all yet).

lang-m:
@fangohr Commit [8bf67c3](https://github.com/ubermag/discretisedfield/commit/8bf67c36fc9f1ec2e2f09215274393862968ce69) changes the `on: push` action trigger to run only on master branch. That avoids the problem of duplicated test runs. Small downside: commits on other branches are tested only after a PR is opened. I think that is no big problem for us. Additionally, I've added `workflow_dispatch` to manually run a test on a different branch.

lang-m:
I like the changes introduced in this PR but I would suggest that we do it for all repos together using the scripts in `devtools` and change directly on `master`. I can do that when you say that this is ready @fangohr (and show you if you are interested).

fangohr:
> I like the changes introduced in this PR but I would suggest that we do it for all repos together using the scripts in `devtools` and change directly on `master`. I can do that when you say that this is ready @fangohr (and show you if you are interested).

Yes - that was my plan: test for one package, and once this is working roll out systematically. I'd like to see it :)

lang-m:
This has been added via [6dc009c](https://github.com/ubermag/discretisedfield/commit/6dc009c97f424b09b2662f2d1cca724a48a2ec37).

PR #153, opened 2022-06-14
Fix region contains

lang-m:
The test `p in region` fails if some directions are inside the region and some within floating point accuracy outside the region.

### DIFF
--- a/discretisedfield/region.py
+++ b/discretisedfield/region.py
@@ -429,9 +429,9 @@ class Region:
             return np.all(
                 np.logical_and(
                     np.less_equal(self.pmin, other)
-                    | np.allclose(self.pmin, other, rtol=tol, atol=tol),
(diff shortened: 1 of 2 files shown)


marijanbeg:
Looks good to me, but I am not sure I understand why replacing `allclose` with `isclose` fixes the issue?

lang-m:
`isclose` returns a bool array that we need instead of a "summary". Example:
```python
>>> a = (0., 1.)
>>> b = (1., 1.-1e-10)
>>> np.less(a, b)
array([True, False])
>>> np.isclose(a, b)
array([False, True])  # allclose would return: False
>>> # For "less than or approx equal" we need element-wise comparison:
>>> np.less(a, b) | np.isclose(a, b)
array([True, True])
# Frist true from np.less, second True from np.isclose
# with allclose we would get: array([True, False])
```

PR #158, opened 2022-07-07
Save subregions to json and reorganise IO.

lang-m:
- Save subregions to json file when saving a field
- Load subregion information from json file (if it exists) when loading a field from disk
- Move IO for `df.Field` to a separate submodule
- Use `pathlib.Path` for io


discretisedfield/field.py
```diff
@@ -2800,170 +2803,30 @@ def write(self, filename, representation="bin8", extend_scalar=False):
...
+        ~discretisedfield.io.field_to_vtk
+        ~discretisedfield.io.field_to_hdf5
 
         """
-        if filename.endswith((".omf", ".ovf", ".ohf")):
-            self._writeovf(
-                filename, representation=representation, extend_scalar=extend_scalar
+        filename = pathlib.Path(filename)
+        if filename.suffix in [".omf", ".ovf", ".ohf"]:
+            io.field_to_ovf(
```
swapneelap, line 2816: Why have you not passed the value of `save_subregions` argument to `write` method to the `io` module functions? For example,
```Python
io.field_to_hdf5(self, filename, save_subregions)
```
Am i missing something?

discretisedfield/tests/test_field.py
```diff
@@ -1725,10 +1729,11 @@ def test_write_read_ovf(self):
 
                     assert f.allclose(f_read)
                     assert f_read.units == "A/m"
+                    assert f.mesh.subregions == f_read.mesh.subregions
```
swapneelap, line 1732: It would be also good to test in all the `test_read_write_*` functions that if one defines `save_subregions=False`, and even if the `subregions` are defined, the `json` file is not being created.

swapneelap (approved):
Added a couple of comments. Everything else looks good to me.

discretisedfield/field.py
lang-m: You are right, this is missing.

PR #169, opened 2022-08-25
Proof-of-concept implementation for `xarray.DataArray` based n-dimensional field [only partly functional]

lang-m:


fangohr:
Method for comparing meshes required first.

PR #173, opened 2022-08-26
Add `from_xarray`, `__array_ufunc__`, and `coordinate_field` methods

swapneelap:


discretisedfield/field_new.py
```diff
@@ -77,8 +77,59 @@ def __init__(
...
+            )
+
+        for i in xa.coords.dims[:-1]:
+            if xa[i].data.size > 1 and not np.allclose(
+                np.diff(xa[i].data), np.diff(xa[i].data).mean()
+            ):
+                raise ValueError(f"Coordinates of {i} must be equally spaced.")
+
+        try:
+            cell = xa.attrs["cell"]
```
marijanbeg, line 98: Do we ensure here that cell makes sense? For instance, there could be some garbage in `cell` attribute passed via xarray.

discretisedfield/field_new.py
```diff
@@ -77,8 +77,59 @@ def __init__(
...
+        except KeyError:
+            if any(len_ == 1 for len_ in xa.data.shape[:-1]):
+                raise KeyError(
+                    "DataArray must have a 'cell' attribute if any "
+                    "of the geometric directions has a single cell."
```
marijanbeg, line 103: geometric -> spatial

discretisedfield/field_new.py
```diff
@@ -77,8 +77,59 @@ def __init__(
...
+                ) from None
```
marijanbeg, line 104: Is `from None` necessary?

discretisedfield/field_new.py
```diff
@@ -77,8 +77,59 @@ def __init__(
...
+        )
+
+        # if any("units" not in xa[i].attrs for i in xa.coords.dims[:-1]):
+        mesh = df.Mesh(p1=p1, p2=p2, cell=cell)  # TODO: Check for units!
+        # else:
+        #     mesh = df.Mesh(
+        #         p1=p1, p2=p2, cell=cell, attributes={"unit": xa["z"].attrs["units"]}
+        #     )
+
+        comp = (
```
marijanbeg, line 125: `comp` -> `components`

discretisedfield/field_new.py
```diff
@@ -361,8 +412,46 @@ def __matmul__(self):  # -> dot
...
         raise NotImplementedError()
 
-    def __array_ufunc__(self):
-        raise NotImplementedError()
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        """Field class support for numpy ``ufuncs``."""
+        # See reference implementation at:
+        # https://numpy.org/doc/stable/reference/generated/numpy.lib.mixins.NDArrayOperatorsMixin.html#numpy.lib.mixins.NDArrayOperatorsMixin
+        for x in inputs:
+            if not isinstance(x, (Field, np.ndarray, numbers.Number)):
```
marijanbeg, line 420: `Field` -> `self.__class__` in case this class is inherited.

discretisedfield/field_new.py
```diff
@@ -361,8 +412,46 @@ def __matmul__(self):  # -> dot
...
+                return NotImplemented
+        out = kwargs.get("out", ())
+        if out:
+            for x in out:
+                if not isinstance(x, Field):
```
marijanbeg, line 425: `Field` -> `self.__class__` in case this class is inherited.

discretisedfield/field_new.py
```diff
@@ -361,8 +412,46 @@ def __matmul__(self):  # -> dot
...
+                    return NotImplemented
+
+        mesh = [x.mesh for x in inputs if isinstance(x, Field)]
```
marijanbeg, line 428: same as above

marijanbeg (commented):
Comments in the code.

discretisedfield/field_new.py
swapneelap: The value of `cell` is passed to `df.Mesh` below. Hence, if it is garbage, it will throw as error from the mesh class. Would you like to have an extra check here?

discretisedfield/field_new.py
swapneelap: I am trying to supress errors from the `xarray.DataArray` which might confuse the user. I can remove it if you like.

discretisedfield/field_new.py
marijanbeg: Good point!

discretisedfield/field_new.py
swapneelap: Sorry, I didn't get your comment. I am passing `comp` to `component` parameter of `df.Mesh` below. Do you want me to change the variable name?

marijanbeg: Yes, I think it is clearer if we call it `components`.

swapneelap: Got it, updated!

discretisedfield/field_new.py
```diff
@@ -77,12 +77,73 @@ def __init__(
...
+    def coordinate_field(cls, mesh):
+        field = cls(mesh, dim=mesh.dim)
+        dim_midpoints_list = [
+            getattr(mesh.midpoints, mesh_dim) for mesh_dim in mesh.dimensions
+        ]
+        dim_points_list = np.meshgrid(*dim_midpoints_list, indexing="ij")
+        for i in range(mesh.dim):
+            field.data.data[..., i] = dim_points_list[i]
+
+        return field
```
marijanbeg, line 146: `return cls(mesh, dim=vdim, value=lambda point: *point)` this might be slow....

discretisedfield/field_new.py
```diff
@@ -77,12 +77,73 @@ def __init__(
...
+        )
 
     @classmethod
-    def coordinate_field(cls):
-        raise NotImplementedError()
+    def coordinate_field(cls, mesh):
+        field = cls(mesh, dim=mesh.dim)
+        dim_midpoints_list = [
+            getattr(mesh.midpoints, mesh_dim) for mesh_dim in mesh.dimensions
+        ]
```
lang-m, line 141: Use `field.data[<dimension_name>].data` instead

PR #174, opened 2022-08-26
Proof-of-concept implementation for xarray.DataArray based n-dimensional field - nd fft functions based on xarray

KevinYanesG:


discretisedfield/field_new.py
```diff
@@ -389,29 +441,57 @@ def laplace(self):
 
     @property
     def fftn(self):
-        raise NotImplementedError()
+        # TODO: I actually think in the last version you were already correcting the
+        #  shift, so we should inclue true_amplitude, true_phase = True
+        ft_xarray = xrft.fft(
+            self.data, dim=[*self.dims], true_amplitude=True, true_phase=True
```
marijanbeg, line 447: Is `[*self.dims]` necessary if `self.dims` is already a tuple?

discretisedfield/field_new.py
```diff
@@ -389,29 +441,57 @@ def laplace(self):
...
+        )
+        return self.from_xarray(ft_xarray)
```
marijanbeg, line 449: `self` or `self.__class__`? `from_xarray` is a class method.

marijanbeg: Same for other returns....

discretisedfield/field_new.py
marijanbeg: Same for other methods.

marijanbeg (commented):
Comments added to the PR review.

KevinYanesG:
Using the `irfft` function creates a problem due to the mesh. `irfft()` calls `_ifft_mesh()` which doesn't store the translational information. Therefore, the original mesh coordinates are not kept. @samjrholt pointed to a code bit that could help us in the `xrft` code base, if I'm not mistaken it should be this: https://github.com/xgcm/xrft/blob/d5a89dceb1ddfc6769fec099c4d141fdeba8a050/xrft/xrft.py#L149-151

As an alternative to this PR we have https://github.com/ubermag/discretisedfield/pull/187, which tries to use directly the `xrft` impplementation, but fails in the `rfftn` method. The error is documented there.

PR #175, opened 2022-08-26
Checks if two fields are defined on the same space

lang-m:
Two functions to check for the same mesh and the same vector dimensions.

### DIFF
--- a/discretisedfield/field_new.py
+++ b/discretisedfield/field_new.py
@@ -233,28 +233,25 @@ class Field:
             raise TypeError(f"Wrong type for {units=}; must be of type str.")
         self._units = units
 
-    def check_same_mesh(self, other):
+    def is_same_mesh(self, other, tolerance_factor=1e-5):
         """Check if two Field objects are defined on the same mesh."""
         if not isinstance(other, self.__class__):
             raise TypeError(f"Object of type {type(other)} not supported.")
         if self.dims != other.dims:
             return False
         for dim in self.dims:
-            print(dim)
-            print(self.data[dim].data)
-            print(other.data[dim].data)
             if len(self.data[dim]) != len(other.data[dim]):
                 # change absolute tolerance to a tolerance relative to the cell size
                 # to take into account the overall scale
                 if not np.allclose(
                     self.data[dim],
                     other.data[dim],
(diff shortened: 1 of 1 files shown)


discretisedfield/field_new.py
```diff
@@ -236,14 +236,28 @@ def units(self, units):
     def check_same_mesh(self, other):
```
marijanbeg, line 236: If `True` or `False` is returned, I would remove `check_` prefix.

discretisedfield/field_new.py
```diff
@@ -236,14 +236,28 @@ def units(self, units):
...
             return False
-        if self.dims != other.dims or self.vdims != other.vdims:
-            return False
-        # for dim is self.dims:
-        #    pass
-        raise NotImplementedError()
+        for dim in self.dims:
+            print(dim)
+            print(self.data[dim].data)
+            print(other.data[dim].data)
```
marijanbeg, line 245: I guess `print` calls were for debugging.

discretisedfield/field_new.py
```diff
@@ -236,14 +236,28 @@ def units(self, units):
...
+                # to take into account the overall scale
+                if not np.allclose(
+                    self.data[dim],
+                    other.data[dim],
+                    atol=self.cell[self.dims.index(dim)] * 1e-5,
+                ):
+                    return False
+        return True
+
+    def check_same_vectorspace(self, other):
```
marijanbeg, line 257: Same as above... I would remove `check_`

discretisedfield/field_new.py
```diff
@@ -236,14 +236,28 @@ def units(self, units):
...
+            print(dim)
+            print(self.data[dim].data)
+            print(other.data[dim].data)
+            if len(self.data[dim]) != len(other.data[dim]):
+                # change absolute tolerance to a tolerance relative to the cell size
+                # to take into account the overall scale
+                if not np.allclose(
+                    self.data[dim],
+                    other.data[dim],
+                    atol=self.cell[self.dims.index(dim)] * 1e-5,
```
marijanbeg, line 252: Does it have to be hardcoded?

PR #176, opened 2022-08-26
Xarray based nd field maths

samjrholt:


discretisedfield/field_new.py
```diff
@@ -327,52 +327,192 @@ def sel(self, *args, **kwargs):
     # mathematical operations
 
     def __abs__(self):
-        raise NotImplementedError()
+        return self.__class__(
+            self.mesh, dim=self.nvdims, value=np.abs(self.data.data), units=self.units
+        )
 
     def __pos__(self):
```
marijanbeg, line 334: This method might work by default, i.e. it might be Python's default behaviour. Removing it and checking if tests pass would be worth doing.

discretisedfield/field_new.py
```diff
@@ -327,52 +327,192 @@ def sel(self, *args, **kwargs):
...
-        raise NotImplementedError()
+        return self
 
     def __neg__(self):
-        raise NotImplementedError()
+        return -1 * self
 
     def __add__(self, other):
         if isinstance(other, self.__class__):
```
marijanbeg, line 341:
```python
if not isinstance(other, self.__class__):
    raise TypeError(...)

...
```

marijanbeg (commented):
LGTM. I would only suggest maybe packing

```python
if isinstance(other, self.__class__):
            if not self.is_same_mesh(other):
                raise ValueError(
                    "To perform this operation both fields must have the same mesh."
                )

            if self.is_vectorfield and other.is_vectorfield:
                if not self.is_same_vectorspace(other):
                    raise ValueError(
                        "To perform this operation both fields must have the same"
                        " vector components."
                    )
```

to a function to avoid code repetition.

discretisedfield/field_new.py
samjrholt: This is delegated to xarray

discretisedfield/field_new.py
samjrholt: @marijanbeg This should be corrected in the latest commit https://github.com/ubermag/discretisedfield/pull/176/commits/2b9b468043c154b21185e61e31f88abe442ca95f

marijanbeg: Danke :)

PR #190, opened 2022-08-28
refactor maths

samjrholt:


fangohr:
I think we should keep the `.average` property, but instead of computing the average issue an Error that explains to use `.mean()` instead of `.average` now. That will help users (and us) to remember what is wrong if we execute older programs.

samjrholt:
@lang-m this PR should deal with all the n dimensional maths apart from calculus and Fourier transforms. All tests should pass apart from 3 ones based on integrals where changing the dot product no longer allows the original functionality but this should be changed when we tackle the integrals.

PR #192, opened 2022-09-02
Proof of concept PR for `numba` guvectorized functions for setting value and norm for a field.

swapneelap:


discretisedfield/field.py
```diff
@@ -335,8 +336,16 @@ def value(self):
 
     @value.setter
     def value(self, val):
-        self._value = val
-        self.array = _as_array(val, self.mesh, self.dim, dtype=self.dtype)
+        if callable(val):
+            nb_value_func = nb.guvectorize(
+                [(nb.float64[:], nb.float64[:])], "(n)->(n)", nopython=True
```
lang-m, line 341: What happens if `nopython=True` is wrong? Is it ignored or does it raise an exception?

lang-m (commented):
I think it would be good to integrate this functionality during the refactoring. The benefit can be significant I guess and just documenting how to do it manually is not very user-friendly (at least for me as a user). Some questions/comments:
- Do we have some actual numbers on the speed-up depending on cell size and complexity?
- The implementation should go to `_as_array` (-> the callable implementation).
- I would suggest to use `try ... except`, try numba first and if it fails use the simple for loop (+ logging, maybe debug level, to know what is going on).

PR #193, opened 2022-09-05
Documentation spell fixing

KevinYanesG:


lang-m (changes requested):

docs/field-holoviews-visualisation.ipynb
```diff
@@ -2062,6 +2062,7 @@
     }
    ],
    "source": [
+    "# Note for web-users: You should use a jupyter server to interact with the plot\n",
```
lang-m, line 2065: Context: The Holoviews plots on the website do not update when dragging the slider (because the need a running Python interpreter). This is a "fundamental problem" and cannot be changed - and that is okay.

@fangohr To avoid possible confusion, Kevin has added a comment to warn users that browse the notebooks on our websites. Do you think that this explanation is understandable for our users (i.e. do they understand when the plot will/will not update and how to "fix" it)?

lang-m (approved):

docs/field-holoviews-visualisation.ipynb
fangohr: Tricky. 

First, it is good to have some comment that explains that the static behaviour is not normal.

Second, we may be able to improve on the concrete wording. "Jupyter server" is perhaps not ideal as it might be understood to imply you have a JupyterHub (which we don't need). Could we talk about a 'live notebook' or a 'active notebook session'? I guess this would work with myBinder?

fangohr: The most extreme (in my view) approach would be to create a video that shows what should be happening - I think Kevin did this for postopus. But in addition to creating work to produce the video, the video may go out of date,  the video needs to be stored, and it may also be a little confusing to the user to understand that the video would be not present when using the notebook directly.

lang-m: I agree 'active notebook session' (or 'live notebook') would be better. I would not (at least not now) create any videos/gifs.

KevinYanesG: Ok, I will make the change probably next monday at the office.

PR #196, opened 2022-09-12
Holoviews callbacks

lang-m:
New implementation for the holoviews-based plotting method allowing us to create interactive plots in `micromagneticdata` without loading all the data first.

Missing/removed
- discretisedfield/xarray for cdim
- dict for n


lang-m:
@KevinYanesG The other PR is ubermag/micromagneticdata#35. Note that the other PR combines several changes. The only relevant bit for you is the holoviews-related code in `abstract_drive.py` (at the very bottom of the file starting with the `hv` property).

discretisedfield/plotting/hv.py
```diff
@@ -55,17 +73,18 @@ def __call__(
          determines the typo of plot in the following order:
```
KevinYanesG, line 73: typo "typo" here :D

discretisedfield/plotting/hv.py
```diff
@@ -165,54 +179,48 @@ def __call__(
...
-                scalar = self.__class__(self.array.sel(comp=scalar_dims)).scalar(
+                key_dims = copy.deepcopy(self.key_dims)
+                if len(scalar_comps) > 1:
+                    key_dims["comp"] = hv_key_dim(scalar_comps, key_dims["comp"].unit)
+                    callback = self.callback
+                else:
+                    # manually remove component 'comp' from returned xarray to avoid
+                    # a drop-down with one element (the out-of-plane component)
+                    def callback(*args, **kwargs):
+                        res = self.callback(*args, **kwargs)
```
KevinYanesG, line 212: What is `self.callback()` here? I'm assuming the inner function is not calling itself recursively and i is still the same `self.callback` as was defined in the `__init__()`. Why do you call the inner function also `callback`?

Could you provide an example of `self.callback`?

KevinYanesG (commented):
Otherwise it looks good to me. Although I need to say, I am far of understanding everything due to the complexity of the code.

discretisedfield/tests/test_field.py
```diff
@@ -128,6 +129,15 @@ def check_field(field):
             assert getattr(field, comp).dim == 1
 
 
+def check_hv(plot, types):
+    # generate the first plot output to have enough data in plot.info
+    hv.renderer("bokeh").get_plot(plot)
+    # assert str(plot) == ''
```
lang-m, line 135:
```suggestion
```

discretisedfield/tests/test_field.py
```diff
@@ -128,6 +129,15 @@ def check_field(field):
...
+    assert sorted(
```
lang-m, line 136:
```suggestion
    # find strings like "    :DynamicMap [comp,z]" or "    :Image    [x,y]"
    # the number of spaces can vary
    assert sorted(
```

PR #199, opened 2022-10-21
Refactor mesh

lang-m:


discretisedfield/tests/test_mesh.py
```diff
@@ -243,6 +286,41 @@ def test_eq(self):
...
+        assert mesh1.allclose(mesh2)
+
+        p1 = (0, 0, 0)
+        p2 = (10 + 1e-12, 10 + 2e-13, 10 + 3e-12)
+        n = (1, 1, 1)
+        atol = 1e-10
+        rtol = 1e-8
+        mesh3 = df.Mesh(p1=p1, p2=p2, n=n)
+
+        assert mesh1.allclose(mesh3, rtol=rtol, atol=atol)
```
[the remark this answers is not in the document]
lang-m: In any case having an additional test with some small deviations within the "built-in" accuracy would be good.

discretisedfield/mesh.py
```diff
@@ -229,6 +238,38 @@ def __init__(
...
+    def bc(self, bc):
+        if bc not in {"neumann", "dirichlet", ""}:
+            for char in bc:
+                if char not in self.region.dims:
+                    raise ValueError(f"Axis {char} is absent in {self.region.dims}.")
+                elif bc.count(char) > 1:
+                    raise ValueError(f"Axis {char} is present more than once.")
+            self._bc = bc
+        else:
+            self._bc = bc
```
lang-m, line 264:
```suggestion

        self._bc = bc
```

discretisedfield/mesh.py
```diff
@@ -229,6 +238,38 @@ def __init__(
...
+    def bc(self):
+        """Boundary condition for the mesh."""
+        return self._bc
+
+    @bc.setter
+    def bc(self, bc):
+        if bc not in {"neumann", "dirichlet", ""}:
+            for char in bc:
+                if char not in self.region.dims:
+                    raise ValueError(f"Axis {char} is absent in {self.region.dims}.")
```
lang-m, line 259: This is based on the assumption that `region.dims` can only have single-letter names (which is not required). We have to think about boundary conditions in more detail and probably need a different syntax. (This is also relevant for @samjrholt )

I would not change anything in this PR.

discretisedfield/mesh.py
```diff
@@ -229,6 +238,38 @@ def __init__(
...
+                elif bc.count(char) > 1:
+                    raise ValueError(f"Axis {char} is present more than once.")
+            self._bc = bc
+        else:
+            self._bc = bc
+
+    @property
+    def cell(self):
+        """The cell dimensions of the mesh."""
+        return self._cell
```
lang-m, line 269: This should be computed on-the-fly. Mesh should only save `n` (to simplify scaling).

discretisedfield/mesh.py
```diff
@@ -229,6 +238,38 @@ def __init__(
...
+
+    @property
+    def n(self):
+        """Number of cells along each dimensions of the mesh."""
```
lang-m, line 273:
```suggestion
        """Number of cells along each dimension of the mesh."""
```

discretisedfield/mesh.py
```diff
@@ -229,6 +238,38 @@ def __init__(
...
+                    raise ValueError(f"Axis {char} is absent in {self.region.dims}.")
+                elif bc.count(char) > 1:
+                    raise ValueError(f"Axis {char} is present more than once.")
+            self._bc = bc
+        else:
+            self._bc = bc
+
+    @property
+    def cell(self):
+        """The cell dimensions of the mesh."""
```
lang-m, line 268: Extend docstring to document return type (and maybe on example).

discretisedfield/mesh.py
lang-m, line 273: Extend docstring to document return type (and maybe on example).

discretisedfield/mesh.py
```diff
@@ -246,6 +287,15 @@ def subregions(self):
 
     @subregions.setter
     def subregions(self, subregions):
+        if not isinstance(subregions, dict):
+            raise TypeError(
+                "Subregions must be a dictionary relating the name of a subregion with"
+                " its region."
+            )
+
+        if not all([isinstance(key, str) for key in subregions]):
```
lang-m, line 304:
```suggestion
        if not all(isinstance(key, str) for key in subregions):
```

discretisedfield/mesh.py
```diff
@@ -380,11 +430,11 @@ def __iter__(self):
             yield self.index2point(index)
 
     def axis_points(self, axis, /):
-        warnings.warn("Deprecated; use `mesh.midpoints` instead.", FutureWarning)
-        return getattr(self.midpoints, axis)
+        warnings.warn("Deprecated; use `mesh.points` instead.", FutureWarning)
+        return getattr(self.points, axis)
```
lang-m, line 442: I would remove this now (maybe raise an AttributeError which points to `points`).

discretisedfield/mesh.py
```diff
@@ -380,11 +430,11 @@ def __iter__(self):
...
 
     @functools.cached_property
```
lang-m, line 444:
```suggestion
    @property
```
We (will) allow in-place scaling and translations and hence cannot cache this anymore.

discretisedfield/mesh.py
```diff
@@ -380,11 +430,11 @@ def __iter__(self):
...
-    def midpoints(self):
+    def points(self):
```
lang-m, line 445: Maybe add an additional `midpoints` and raise an `AttributeError` pointing to `points`.

discretisedfield/mesh.py
```diff
@@ -909,6 +900,15 @@ def plane(self, *args, n=None, **kwargs):
         return plane_mesh
 
     def __or__(self, other):
+        """Depricated method to check if meshes are aligned: use ``is_aligned``"""
```
lang-m, line 911:
```suggestion
        #Deprecated method to check if meshes are aligned: use ``is_aligned``
```
I would remove the docstring to remove the operator from the documentation.

discretisedfield/mesh.py
```diff
@@ -909,6 +900,15 @@ def plane(self, *args, n=None, **kwargs):
...
+
+        warnings.warn(
+            "Bitwise OR (|) operator is depricated; please use is_aligned",
```
lang-m, line 914:
```suggestion
            "Bitwise OR (|) operator is deprecated; please use is_aligned",
```

discretisedfield/mesh.py
```diff
@@ -917,18 +917,32 @@ def __or__(self, other):
```
lang-m, line 925: Does point 1 mean `same discretisation cell size`?

discretisedfield/mesh.py
```diff
@@ -917,18 +917,32 @@ def __or__(self, other):
...
         bool
 
             ``True`` if meshes are aligned, ``False`` otherwise.
 
+        Raises
+        ------
+        TypeError
+
+            If ``other`` argument is not of type ``discretisedfield.Mesh`` or if
+            ``tolerance`` argument is not of type either ``float`` or ``int``.
```
lang-m, line 952:
```suggestion
            ``tolerance`` argument is not of type ``float`` or ``int``.
```

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
             description=description,
             disabled=False,
         )
+
+    def allclose(self, other, rtol=1e-05, atol=1e-08):
+        """Check if the mesh is close enough to the other based on a tolerance.
+
+        The midpoints (``discretisedfield.Mesh.points``) along all the dimensions for
```
lang-m, line 1559: Is "midpoints" good (now that it is called points)?

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+        both the mesh are compared utilizng ``numpy.allclose``. The value of relative
```
lang-m, line 1560:
```suggestion
        both meshes are compared using ``numpy.allclose``. The value of relative
```

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+        tolerance (``rtol``) and absolute tolerance (``atol``) are passed on to
+        ``numpy.allclose`` for the comparison. If the midpoints along all the
```
lang-m, line 1562: midpoints (?)

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+        dimensions are close enough simultaneously, the method returns ``True``
```
lang-m, line 1563:
```suggestion
        dimensions are close enough, the method returns ``True``
```

discretisedfield/mesh.py
lang-m, line 1562:
```suggestion
        ``numpy.allclose`` for the comparison. If the midpoints along all
```

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
             description=description,
             disabled=False,
         )
+
+    def allclose(self, other, rtol=1e-05, atol=1e-08):
```
lang-m, line 1556: We have to think about better defaults. The default `atol` is much bigger that the typical cell size (order 1e-9).

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+        bool
+
+            ``True`` if other mesh is close enough, otherwise ``False``.
+
+        Raises
+        ------
+        TypeError
+
+            If the ``other`` argument is not of type ``discretisedfield.Mesh`` or if
+            ``rtol`` and ``atol`` arguments are not of type either ``float`` or ``int``.
```
lang-m, line 1593:
```suggestion
            ``rtol`` and ``atol`` arguments are not of type ``float`` or ``int``.
```

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+        >>> p2 = (20e-9, 20e-9, 20e-9)
+        >>> n = (10, 10, 10)
+        >>> mesh1 = df.Mesh(p1=p1, p2=p2, n=n)
+        ...
+        >>> p1 = (0, 0, 0)
+        >>> p2 = (20e-9 + 1.2e-12, 20e-9 + 1e-13, 20e-9 + 2e-12)
+        >>> n = (10, 10, 10)
+        >>> mesh2 = df.Mesh(p1=p1, p2=p2, n=n)
+        ...
+        >>> mesh1.allclose(mesh2, rtol=1e-6, atol=1e-11)
```
lang-m, line 1611: I would prefer if the "standard" example would work without user-defined rtol and atol.

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+                f"Expected argument of type discretisedfield.Mesh but got {type(other)}"
+            )
+
+        if self.region.dims != other.region.dims:
+            raise ValueError("The mesh dimensions do not match.")
+
+        if (not isinstance(rtol, (int, float))) or (not isinstance(atol, (int, float))):
+            raise TypeError(
+                "Expected both rtol and atol to be either int or float but got"
+                f" {type(rtol)} and {type(atol)} respectively"
```
lang-m, line 1629:
```suggestion
                f" {type(rtol)} and {type(atol)}, respectively."
```

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+            [
+                np.allclose(
+                    getattr(self.points, dim),
+                    getattr(other.points, dim),
+                    rtol=rtol,
+                    atol=atol,
+                )
+                for dim in self.region.dims
+            ]
+        )
```
lang-m, line 1642: The brackets inside `all(...)` can be omitted (should be done locally to run black before the commit).

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+
+    def coordinate_field(self):
+        """Create a field whose values are the mesh coordinates.
+
+        This method can be used to create a vector field with values equal to the
+        coordinates of the cell midpoints. The result is equivalent to a field created
```
lang-m, line 1648: midpoints (?)

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+        with the following code:
+
+        .. code-block::
+
+            mesh = df.Mesh(...)
+            df.Field(mesh, dim=3, value=lambda point: point)
```
lang-m, line 1654:
```suggestion
            df.Field(mesh, dim=mesh.region.ndim, value=lambda point: point)
```
(to make it generic for the future)

discretisedfield/mesh.py
```diff
@@ -1505,3 +1544,144 @@ def axis_selector(self, *, widget="dropdown", description="axis"):
...
+
+        >>> manually = df.Field(mesh, dim=3, value=lambda point: point)
+        >>> cfield.allclose(manually)
+        True
+
+        """
+
+        field = df.Field(self, dim=self.region.ndim)
+        for i, dim in enumerate(self.region.dims):
+            field.array[..., i] = getattr(self.points, dim).reshape(
```
lang-m, line 1691:
```suggestion
            points = self.points  # avoid re-computing points
            field.array[..., i] = getattr(points, dim).reshape(
```

discretisedfield/mesh.py
```diff
@@ -187,22 +182,42 @@ def __init__(
...
 
-        if cell is not None and n is None:
-            self.cell = tuple(cell)
-            n = np.divide(self.region.edges, self.cell).round().astype(int)
-            self.n = dfu.array2tuple(n)
+        if cell is None and n is None:
+            raise ValueError("Both cell and n cannot be None.")
+        elif cell is not None and n is None:
+            if not isinstance(cell, Iterable):
+                raise TypeError("Cell must be an Iterable.")
```
lang-m, line 201: I think `Iterable` is too generic. I would only allow array-like types (tuple, list, np.ndarray). I am not sure if there is a generic test for this.

discretisedfield/mesh.py
```diff
@@ -187,22 +182,42 @@ def __init__(
...
+            self._cell = tuple(cell)
+            if len(self._cell) != self.region.ndim:
```
lang-m, line 203:
```suggestion
            if len(cell) != self.region.ndim:
```

discretisedfield/mesh.py
```diff
@@ -187,22 +182,42 @@ def __init__(
...
+                    "The values of cell must be of type either int, float or complex"
+                )
+            n = np.divide(self.region.edges, self._cell).round().astype(int)
+            self._n = dfu.array2tuple(n)
         elif n is not None and cell is None:
-            self.n = tuple(n)
-            cell = np.divide(self.region.edges, self.n).astype(float)
-            self.cell = dfu.array2tuple(cell)
+            if not isinstance(n, Iterable):
+                raise TypeError("n must be an Iterable.")
```
lang-m, line 213: No `Iterable`, array-like instead.

discretisedfield/mesh.py
```diff
@@ -187,22 +182,42 @@ def __init__(
...
+            self._n = tuple(n)
+            if len(self._n) != self.region.ndim:
```
lang-m, line 215:
```suggestion
            if len(n) != self.region.ndim:
```

discretisedfield/mesh.py
```diff
@@ -187,22 +182,42 @@ def __init__(
...
+                raise ValueError("n must have same dimensions as the region.")
+            elif not all([isinstance(i, Integral) for i in self._n]):
+                raise TypeError("The values of n must be of integer type.")
+            cell = np.divide(self.region.edges, self._n).astype(float)
+            self._cell = dfu.array2tuple(cell)
```
lang-m, line 220: No need to compute `cell` here.

discretisedfield/mesh.py
```diff
@@ -187,22 +182,42 @@ def __init__(
...
-            self.n = tuple(n)
-            cell = np.divide(self.region.edges, self.n).astype(float)
-            self.cell = dfu.array2tuple(cell)
+            if not isinstance(n, Iterable):
+                raise TypeError("n must be an Iterable.")
+            self._n = tuple(n)
+            if len(self._n) != self.region.ndim:
+                raise ValueError("n must have same dimensions as the region.")
+            elif not all([isinstance(i, Integral) for i in self._n]):
+                raise TypeError("The values of n must be of integer type.")
```
lang-m, line 218: Are we checking that the values of `n` are positive?

discretisedfield/mesh.py
```diff
@@ -219,6 +234,8 @@ def __init__(
             )
             raise ValueError(msg)
 
+        if not isinstance(bc, str):
+            raise TypeError("Value of bc must be a string.")
         self.bc = bc.lower()
```
lang-m, line 239: All checks (and conversion to `lower`) should be in `bc.setter`.

discretisedfield/mesh.py
```diff
@@ -187,22 +182,42 @@ def __init__(
         attributes={"unit": "m"},
     ):
         # TODO NO MUTABLE DEFAULT
-        if region is not None and p1 is None and p2 is None:
-            self.region = region
+        if region is None and p1 is None and p2 is None:
+            raise ValueError("p1, p2, and region cannot be None simultaneously.")
```
lang-m, line 186: Do we need two separate checks for "all None", which would anyway fail if e.g. only region and p1 are not None, and the `else` (line 193)?

discretisedfield/mesh.py
samjrholt: I think this is good for now and we can think about `bc` in more detail in the future.

discretisedfield/mesh.py
```diff
@@ -229,6 +238,38 @@ def __init__(
         if "isplane" not in attributes.keys():
             self.attributes["isplane"] = False
 
+    @property
+    def bc(self):
+        """Boundary condition for the mesh."""
```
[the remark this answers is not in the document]
samjrholt: I agree with adding the part of the documentation from the mesh class to do with `bc` here

discretisedfield/mesh.py
```diff
@@ -187,22 +182,42 @@ def __init__(
...
-            self.n = dfu.array2tuple(n)
+        if cell is None and n is None:
+            raise ValueError("Both cell and n cannot be None.")
+        elif cell is not None and n is None:
+            if not isinstance(cell, Iterable):
+                raise TypeError("Cell must be an Iterable.")
+            self._cell = tuple(cell)
+            if len(self._cell) != self.region.ndim:
+                raise ValueError("The cell must have same dimensions as the region.")
+            elif not all([isinstance(i, (Number, complex)) for i in self._cell]):
```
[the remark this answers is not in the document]
samjrholt: So far we have not had a use case for a complex valued mesh so I do not think it will be a use case that is common at all. Hence, I'm happy not to allow it if it makes our lives easier. However, where does the difficulty come from in supporting these complex meshes? As we are based on `numpy` aren't most of the operations deal with complex numbers automatically.

discretisedfield/tests/test_mesh.py
```diff
@@ -179,11 +179,54 @@ def test_cell_greater_than_domain(self):
...
+            df.Mesh(p1=p1, p2=p2, cell=cell, n=n)
 
     def test_bc(self):
-        raise NotImplementedError
+        p1 = (0, 0, 0)
+        p2 = (20e-9, 20e-9, 20e-9)
+        region = df.Region(p1=p1, p2=p2, dims=["x", "y", "z"])
+        cell = (2e-9, 4e-9, 1e-9)
+
+        for bc in ["x", "y", "yz", "zx", "xyz", "Neumann", "dirichlet"]:
```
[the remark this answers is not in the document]
samjrholt: I think we have been using `neumann` and `dirichlet` both without capitalisation so far.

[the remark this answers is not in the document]
lang-m: Boundary conditions are internally converted to lower case. So checking both is actually good.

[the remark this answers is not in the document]
samjrholt: Okay ideal :) I didn't realise

discretisedfield/mesh.py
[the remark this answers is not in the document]
lang-m: I guess it could cause problems with plotting. Furthermore, proper consistency checks between pmin/pmax and cell could be complicated (e.g. real-valued pmin/pmax with purely imaginary cells do not work, complex-valued pmin/pmax and cell must all consistently have non-zero (or zero) real/imaginary parts, ...).

discretisedfield/tests/test_mesh.py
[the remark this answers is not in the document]
swapneelap: No perticular reason. I was thinking that the list should be a good representation of full permuatation and combination of `x`, `y`, and `z`. I can add the full list if you like.

discretisedfield/tests/test_mesh.py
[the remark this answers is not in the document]
swapneelap: When you say "built-in", it means the tolerance defined in `Region` or the default value of `rtol` and `atol` or someting else?

discretisedfield/mesh.py
swapneelap: Yes, good point. It would be good to discuss this,

discretisedfield/mesh.py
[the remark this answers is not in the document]
swapneelap: I agree. I'll move the class docstrings to the properties.

discretisedfield/tests/test_mesh.py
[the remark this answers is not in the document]
lang-m: I would maybe add all.

discretisedfield/mesh.py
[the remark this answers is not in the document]
swapneelap: I'll remove the complex check then.

discretisedfield/mesh.py
swapneelap: The value of `cell` is a genexp in [`io.field_from_ovf`](https://github.com/ubermag/discretisedfield/blob/c8ff5594418189801d1f750e57e22a32e239e09a/discretisedfield/io/ovf.py#L246), and hence, raises `TypeError`. If you like I can change this in the `io` module.

lang-m: That would be good.

discretisedfield/mesh.py
```diff
@@ -187,22 +182,42 @@ def __init__(
...
+                raise TypeError(
+                    "The values of cell must be of type either int, float or complex"
+                )
+            n = np.divide(self.region.edges, self._cell).round().astype(int)
+            self._n = dfu.array2tuple(n)
```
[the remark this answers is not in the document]
swapneelap: Should I replace `array2tuple` with just `tuple`?

discretisedfield/mesh.py
```diff
@@ -282,57 +282,60 @@ def subregions(self):
...
-
-        # Check if subregions are aligned with the mesh
-        for key, value in subregions.items():
-            # Is the subregion in the mesh region?
-            if value not in self.region:
-                msg = f"Subregion {key} is not in the mesh region."
-                raise ValueError(msg)
+        if subregions is None:
+            self._subregions = None
+        else:
```
lang-m, line 296: I maybe was not precise enough here. I would keep the empty dict for the subregions if there are no subergions. However, the default value in `__init__` should not be mutable. So something like:
```python
if subregions is None:
    subregions = {}

File: discretisedfield/tests/test_field.py [hunk 40]
```diff
@@ -1671,7 +1671,7 @@ def test_write_read_ovf(self, tmp_path):
 
                 assert f.allclose(f_read)
                 assert f_read.units == "A/m"
-                assert f_read.mesh.subregions == {}
+                assert f_read.mesh.subregions is None
```
lang-m, line 1674: Nothing should change in the tests.

discretisedfield/mesh.py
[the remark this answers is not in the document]
swapneelap:
```python
return array.item() if array.size == 1 else tuple(array.tolist())
```
Some tests might fail if we replace this with a tuple?

discretisedfield/mesh.py
```diff
@@ -265,7 +261,8 @@ def bc(self, bc):
     @property
     def cell(self):
         """The cell dimensions of the mesh."""
-        return self._cell
+        cell = np.divide(self.region.edges, self.n).astype(float)
+        return dfu.array2tuple(cell)
```
lang-m, line 265: I would return a numpy array and not a tuple.

swapneelap: What about `n`?

discretisedfield/mesh.py
lang-m: We can then presumably avoid the `else` because an empty dict should not cause any problems.

discretisedfield/mesh.py
lang-m: Not sure

discretisedfield/mesh.py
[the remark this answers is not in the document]
lang-m: Maybe a numpy array for `n` would indeed be the best option?

discretisedfield/mesh.py
lang-m: A numpy array would probably be best.

discretisedfield/mesh.py
swapneelap: I do not fully understand what you mean?

swapneelap:
@lang-m, the `Mesh.__repr__` needs to be updated as `n` is a `numpy.ndarray` now. Can you take care of this?

lang-m:
I can update the representation strings.

discretisedfield/mesh.py
lang-m: The explanation above (l. 916) says `... same discretisation cells` which could be interpreted as "both meshes have the same region and cell size -> the same cell**s**. We want to say that both meshes have "the same cell size". So maybe changing `cells` to `cell size` would be better?

swapneelap:
@lang-m and @samjrholt. I think the branch is good to be merged with `refactor_master`. Let me know what you think?

lang-m:
I will review the changes and let you know when I am done.

discretisedfield/mesh.py
```diff
@@ -229,6 +233,77 @@ def __init__(
...
+                if char not in self.region.dims:
+                    raise ValueError(f"Axis {char} is absent in {self.region.dims}.")
+                elif bc.count(char) > 1:
+                    raise ValueError(f"Axis {char} is present more than once.")
+
+        self._bc = bc
+
+    @property
+    def cell(self):
+        """The cell dimensions of the mesh.
```
lang-m, line 272: Would `cell size` be better?
```suggestion
        """The cell size of the mesh.
```

discretisedfield/mesh.py
```diff
@@ -229,6 +233,77 @@ def __init__(
...
+        numpy.ndarray
+
+            A numpy array representing number of discretisation cells along respective
+            axes.
+        """
+        return self._n
+
+    @property
+    def region(self):
+        """Region over which the mesh is defined.
```
lang-m, line 297: Maybe (not sure)

```suggestion
        """Region on which the mesh is defined.
```

discretisedfield/mesh.py
```diff
@@ -240,12 +315,24 @@ def subregions(self):
             A dictionary defining subregions in the mesh. The keys of the
             dictionary are the region names (``str``) as valid Python variable
             names, whereas the values are ``discretisedfield.Region`` objects.
-            Defaults to an empty dictionary.
+            Defaults to None.
```
lang-m, line 318:
```suggestion
```

For the property we can not have a default value.

discretisedfield/mesh.py
```diff
@@ -40,12 +34,6 @@ class Mesh:
     discretisation cells ``n`` in all three dimensions. Either ``cell`` or
     ``n`` can be passed, not both.
```
lang-m, line 36: Should we briefly mention the attribute `bc` here? ("can be used to define different types of boundary conditions; string")

lang-m (commented):
A few last points. Afterwards it can be merged.

PR #203, opened 2022-10-30
Refactor region: scale and translate

lang-m:


discretisedfield/region.py
```diff
@@ -640,16 +640,16 @@ def multiplier(self):
...
+        """Scale the region.
 
         It can be applied only between ``discretisedfield.Region`` and
         ``numbers.Real``. The result is a region whose ``pmax`` and ``pmin``
         are multiplied by ``other``.
 
         Parameters
         ----------
-        other : numbers.Real
+        other : array-like or numbers.Real
```
samjrholt, line 652: Currently we are allowing `pmin` and `pmax` to be `numbers.Number` so I think we could allow factor to be the same type

discretisedfield/region.py
```diff
@@ -640,16 +640,16 @@ def multiplier(self):
...
         return uu.si_max_multiplier(self.edges)
 
-    def __mul__(self, other):
-        """Binary ``*`` operator.
+    def __mul__(self, factor):
+        """Scale the region.
 
         It can be applied only between ``discretisedfield.Region`` and
         ``numbers.Real``. The result is a region whose ``pmax`` and ``pmin``
         are multiplied by ``other``.
```
samjrholt, line 648: factor

discretisedfield/region.py
samjrholt, line 652: factor

discretisedfield/region.py
```diff
@@ -746,7 +736,59 @@ def __truediv__(self, other):
...
+    def __rsub__(self, vector):
+        return self - vector
+
+    def translate(self, vector):
+        """Translate the region."""
+        self._check_operand(vector, False, "method translate")
+        return self._apply_operator(vector, np.add)
+
+    def _check_operand(self, other, allow_scalar, operation):
+        if allow_scalar and isinstance(other, numbers.Real):
```
samjrholt, line 769: See earlier comment regarding `numbers.Number`

discretisedfield/region.py
```diff
@@ -746,7 +736,59 @@ def __truediv__(self, other):
...
+        elif not isinstance(other, (tuple, list, np.ndarray)):
+            raise TypeError(f"Unsupported type {type(other)} for {operation}.")
+        elif len(other) != self.ndim:
+            raise ValueError(
+                f"Wrong length for array-like argument: {len(other)}; expected length"
+                f" {len(self.pmin)}."
+            )
+
+        for elem in other:
+            if not isinstance(elem, numbers.Real):
```
samjrholt, line 780: See earlier comment regarding `numbers.Number`

discretisedfield/region.py
```diff
@@ -746,7 +736,59 @@ def __truediv__(self, other):
...
+    def __radd__(self, vector):
+        return self + vector
+
+    def __sub__(self, vector):
+        """Translate the region."""
+        self._check_operand(vector, False, "operator -")
+        return self._apply_operator(vector, np.subtract)
+
+    def __rsub__(self, vector):
+        return self - vector
```
samjrholt, line 761: Is this correct? Should it be something like ` -1 * self + vector`?

discretisedfield/tests/test_region.py
```diff
@@ -292,12 +292,117 @@ def test_mul_truediv(self):
...
+        assert isinstance(region, df.Region)
+        assert np.allclose(region.pmin, (-50e-9, -50e-9, 0))
+        assert np.allclose(region.pmax, (50e-9, 50e-9, 20e-9))
+        assert np.allclose(region.edges, (100e-9, 100e-9, 20e-9))
+
         with pytest.raises(TypeError):
             res = region * region
 
         with pytest.raises(TypeError):
             res = 5 / region
```
samjrholt, line 311: Do we want a test for multiplication with iterables?

discretisedfield/region.py
lang-m: You are right, this is not correct.

Maybe we should not implement `__radd__` and `__rsub__` (and maybe `__rmul__`). Thoughts @samjrholt ?

samjrholt (approved):

PR #220, opened 2022-11-24
Rename: field.units -> field.unit

lang-m:
Renames: units -> unit

We should maybe also rename the input argument for `to_xarray`.


lang-m:
@swapneelap This is now ready for review.

swapneelap (approved):
Looks good to me.

PR #225, opened 2022-11-29
Refactor value

samjrholt:


discretisedfield/field.py
```diff
@@ -272,29 +259,23 @@ def value(self):
...
-        >>> field.array[0, 0, 0, :] = (0, 0, 0)
-        >>> field.value
-        array(...)
-        >>> field.value.shape
-        (2, 2, 1, 3)
+        >>> field.update_field_values(value_function)
+        >>> field.array[0, 0, 0, :]
+        array([0., 0., 1.])
+        >>> field.array[1, 0, 0, :]
+        array([ 0.,  0., -1.])
```
lang-m, line 278: Would it be better to use `__call__` instead of directly accessing the numpy array?

discretisedfield/field.py
```diff
@@ -330,16 +311,8 @@ def value(self):
...
-        if np.array_equal(self.array, value_array):
-            return self._value
-        else:
-            return self.array
-
-    @value.setter
-    def value(self, val):
-        self._value = val
-        self.array = _as_array(val, self.mesh, self.nvdim, dtype=self.dtype)
+        self._value = value
```
lang-m, line 314: Do we need `_value` anywhere? It would be good if we could remove this line.

discretisedfield/tests/test_field.py
```diff
@@ -179,19 +179,20 @@ class WrongType:
...
-                    *f.mesh.n,
-                    f.nvdim,
+            f.update_field_values(
+                np.ones(
+                    (
+                        *f.mesh.n,
+                        f.nvdim,
+                    )
                 )
             )
```
lang-m, line 189:
```suggestion
            f.update_field_values(np.ones((*f.mesh.n, f.nvdim)))
```

That should be short enough to fit in one line. No idea why black did not change this.

discretisedfield/tests/test_field.py
```diff
@@ -368,19 +366,13 @@ def test_norm(self):
         mesh = df.Mesh(p1=(0, 0, 0), p2=(10, 10, 10), cell=(5, 5, 5))
         f = df.Field(mesh, nvdim=3, value=(2, 2, 2))
 
-        assert np.all(f.norm.value == 2 * np.sqrt(3))
         assert np.all(f.norm.array == 2 * np.sqrt(3))
```
lang-m, line 369: Could we use f.norm.update_field_values here?

discretisedfield/tests/test_field.py
```diff
@@ -368,19 +366,13 @@ def test_norm(self):
...
         assert np.all(f.array == 2)
```
lang-m, line 370: A bit off-topic: I think `np.allclose` would be better.

lang-m (changes requested):
Looks good, most of my comments are somewhat optional.

The only think I find important is (if possible) removing `self._value` (see my comment there).

discretisedfield/tests/test_field.py
samjrholt: Unfortunately it does not work. If we want this to work I think it would be worth having a chat about ideas on how to implement this.

lang-m: I would suggest we talk and then address this separately.

lang-m (approved):

PR #226, opened 2022-11-29
Refactor mean

samjrholt:
Need `mesh.sel`.


discretisedfield/field.py
```diff
@@ -662,12 +663,82 @@ def mean(self, axis=None):
...
+
+        >>> import discretisedfield as df
+        ...
+        >>> p1 = (0, 0, 0)
+        >>> p2 = (5, 5, 5)
+        >>> cell = (1, 1, 1)
+        >>> mesh = df.Mesh(p1=p1, p2=p2, cell=cell)
+        ...
+        >>> field = df.Field(mesh=mesh, nvdim=3, value=(0, 0, 1))
+        >>> field.mean(direction=['x', 'y'])((2.5, 2.5, 0.5))
```
lang-m, line 689: Do we want a list of strings or just one string with multiple elements such as `'xy'`? I know that the latter would cause problems with multi-character direction names. However, we currently have the same limitations in other parts of the code.

discretisedfield/field.py
```diff
@@ -662,12 +663,82 @@ def mean(self, axis=None):
...
 
-        return np.stack(
-            [self.array[..., i].mean(axis=axis) for i in range(self.nvdim)], axis=-1
+        return self.__class__(
+            self.mesh.plane(direction),  # mesh sel method
+            nvdim=self.nvdim,
+            value=self.array.mean(axis=axis),
+            vdims=self.vdims,
+            unit=self.unit,
         )
```
lang-m, line 742: This must be some old code? It will never be executed and can be removed.

discretisedfield/tests/test_field.py
```diff
@@ -495,6 +495,42 @@ def test_mean(self):
         f = df.Field(mesh, nvdim=3, value=(0, 1, 2))
         assert np.allclose(f.mean(), (0, 1, 2))
 
+        # Test with direction
+        assert np.allclose(f.mean(direction=None), f.mean())
```
lang-m, line 499: Not sure if this is necessary.

lang-m (changes requested):
The new `mean` replaces the old `project` method. I would suggest that we remove that method in this PR (just keeping an AttributeError mentioning the new method).

discretisedfield/field.py
```diff
@@ -662,12 +663,82 @@ def mean(self, axis=None):
...
+            if not all(d in self.mesh.region.dims for d in direction):
+                raise ValueError(
+                    f"Invalid direction. Directions must be in {self.mesh.region.dims}."
+                )
+            elif len(direction) != len(set(direction)):
+                raise ValueError("Duplicate directions are not allowed.")
+
+            if sorted(direction) == sorted(self.mesh.region.dims):
+                return self.array.mean(axis=tuple(range(self.mesh.region.ndim)))
+            else:
```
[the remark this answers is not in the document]
samjrholt: Add comments to code

PR #230, opened 2022-11-30
Change return value of field.__call__ to np.ndarray

lang-m:


lang-m:
A few doctests fail. I did not modify doctests for methods where we have other PRs (planned) to avoid redoing the same thing or having to resolve merge conflicts. I would suggest fixing the remaining tests later.

PR #237, opened 2022-12-05
Lapacian bug

samjrholt:


discretisedfield/field.py
```diff
@@ -2021,25 +2020,17 @@ def diff(self, direction, order=1):
                     padded_array, self.mesh.cell[direction_idx], axis=direction_idx
```
lang-m, line 2020: Regarding the comment two lines above: Isn't this a case where we can only compute forward/backward differences?

discretisedfield/field.py
```diff
@@ -2021,25 +2020,17 @@ def diff(self, direction, order=1):
...
-                # The derivative is computed using the central difference
-                # with forward/backward difference at the boundaries.
                 if self.mesh.bc == "":
                     # Pad with specific values so that the same finite difference
                     # stencil can be used across the whole array
                     # central difference = forward difference
-                    # f(1) + f(-1) - 2 f(0) = f(2) + f(0) - 2 f(1)
-                    # f(-1) = f(2) - 3 f(1) + 3f(0)
-
+                    # 0.5 * f(1) - 0.5 * f(-1) = - 0.5 f(2) + 2 f(1) - 1.5 f(0)
```
lang-m, line 2027: Understanding these coefficients (or where they come from) could be a bit difficult. Mabye a reference to findiff (and mentioning why we cannot use it directly) could be useful?

discretisedfield/field.py
```diff
@@ -2021,25 +2020,17 @@ def diff(self, direction, order=1):
...
+                    # f(-1) = f(2) - 3 f(1) + 3 f(0)
```
lang-m, line 2028: Maybe also explain in more detail how this translates in to padding the array.

discretisedfield/field.py
```diff
@@ -2048,21 +2039,68 @@ def pad_fun(vector, pad_width, iaxis, kwargs):
...
+                    # Pad with specific values so that the same finite difference
+                    # stencil can be used across the whole array
+                    # central difference = forward difference
+                    # f(1) + f(-1) - 2 f(0) = f(2) + f(0) - 2 f(1)
+                    # f(-1) = f(2) - 3 f(1) + 3f(0)
+
+                    def pad_fun(vector, pad_width, iaxis, kwargs):
+                        if iaxis == direction_idx:
+                            vector[0] = vector[3] - 3 * vector[2] + 3 * vector[1]
+                            vector[-1] = vector[-4] - 3 * vector[-3] + 3 * vector[-2]
```
lang-m, line 2063: Should have asked this before: When do we need this, i.e. for how many cells? The combination of `n < 4` and index `3` looks "strange".

lang-m (commented):
I cannot comment on the numerical details. I guess you have used findiff to compute the factors.

It would be good to somewhere explain why we do not use findiff.

discretisedfield/field.py
samjrholt: Hopefully I won't explain it here as it will be be good test to see if you can now tell from updated documentation.

lang-m: Makes more sense now. Last question: is this accuracy 1 (instead of the "normal" 2)? Mentioning this would be useful.

lang-m (approved):

PR #238, opened 2022-12-05
Refactor io

lang-m:
- Field io as mixin class
- new hdf5 format


lang-m:
- [ ] additional tests for hdf5 without subregions

samjrholt:
Also the `test_save_load_subregions` test seems to be failing.

lang-m:
`test_save_load_subregions` fails only on Windows.

samjrholt (approved):

PR #251, opened 2022-12-09
refactor(Field): `to_xarray` and `from_xarray` methods

swapneelap:
Refactor both the methods to make use of generalised region.dims and discretisedfield being n-dimensional.

### DIFF
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -3417,7 +3417,7 @@ class Field(_FieldIO):
             msg = "Unit argument must be a string."
             raise TypeError(msg)
 
-        axes = ["x", "y", "z"]
+        axes = self.mesh.region.dims
 
         data_array_coords = {axis: getattr(self.mesh.points, axis) for axis in axes}
 
@@ -3442,6 +3442,7 @@ class Field(_FieldIO):
                 cell=self.mesh.cell,
                 pmin=self.mesh.region.pmin,
                 pmax=self.mesh.region.pmax,
+                nvdim=self.nvdim,
             ),
         )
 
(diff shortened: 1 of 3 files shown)


discretisedfield/field.py
```diff
@@ -3553,29 +3563,29 @@ def from_xarray(cls, xa):
                     "DataArray must have a 'cell' attribute if any "
```
lang-m, line 3563: Two lines above there is a hard-coded `shape[:3]`. I assume this has to be generalised?

discretisedfield/field.py
```diff
@@ -3553,29 +3563,29 @@ def from_xarray(cls, xa):
...
         p2 = (
             xa.attrs["pmax"]
             if "pmax" in xa.attrs
-            else [xa[i].values[-1] + c / 2 for i, c in zip("xyz", cell)]
+            else [xa[i].values[-1] + c / 2 for i, c in zip(dims_list, cell)]
         )
 
-        if any("units" not in xa[i].attrs for i in "xyz"):
+        if any("units" not in xa[i].attrs for i in dims_list):
             mesh = df.Mesh(p1=p1, p2=p2, cell=cell)
```
lang-m, line 3580:
```suggestion
            region = df.Region(p1=p1, p2=p2, dims=dims_list)
            mesh = df.Mesh(region=region, cell=cell)
```

discretisedfield/field.py
```diff
@@ -3409,14 +3409,14 @@ def to_xarray(self, name="field", unit=None):
...
-        axes = ["x", "y", "z"]
+        axes = self.mesh.region.dims
 
         data_array_coords = {axis: getattr(self.mesh.points, axis) for axis in axes}
 
         geo_units_dict = dict(zip(axes, self.mesh.region.units))
 
         if self.nvdim > 1:
-            data_array_dims = axes + ["comp"]
+            data_array_dims = axes + ("comp",)
```
lang-m, line 3419: We should maybe consider renaming `comp` to `vdims`.

lang-m (changes requested):

PR #255, opened 2022-12-12
Tests mesh

samjrholt:


discretisedfield/tests/test_mesh.py
```diff
@@ -834,52 +1028,108 @@ def test_getitem():
...
-    padded_mesh = mesh.pad({"z": (2, 3)})
-    assert np.allclose(padded_mesh.region.pmin, (-1, 2, 2), atol=0)
-    assert np.allclose(padded_mesh.region.pmax, (5, 9, 10), atol=0)
-    assert np.all(padded_mesh.n == (6, 7, 8))
-
-    padded_mesh = mesh.pad({"x": (1, 1), "y": (1, 1), "z": (1, 1)})
-    assert np.allclose(padded_mesh.region.pmin, (-2, 1, 3), atol=0)
-    assert np.allclose(padded_mesh.region.pmax, (6, 10, 8), atol=0)
-    assert np.all(padded_mesh.n == (8, 9, 5))
+    for dim in mesh.region.dims:
```
fangohr, line 1064:
```suggestion
    for dim in mesh.region.dims:  # adding one padding on either side, and for each dimension
```

discretisedfield/tests/test_mesh.py
```diff
@@ -834,52 +1028,108 @@ def test_getitem():
...
+            temp[idx] = temp[idx] + cell[idx]
+        assert np.allclose(padded_mesh.region.pmax, temp, atol=0)
+        temp = mesh.n.copy()
+        if mesh.region.ndim == 1:
+            temp = temp + 2
+        else:
+            temp[idx] = temp[idx] + 2
+        assert np.all(padded_mesh.n == temp)
+
+    for dim in mesh.region.dims:
```
fangohr, line 1087:
```suggestion
    for dim in mesh.region.dims:  # adding padding only on pmax side, and for each dimension
```

discretisedfield/tests/test_mesh.py
```diff
@@ -834,52 +1028,108 @@ def test_getitem():
...
+            temp[idx] = temp[idx] + cell[idx]
+        assert np.allclose(padded_mesh.region.pmax, temp, atol=0)
+        temp = mesh.n.copy()
+        if mesh.region.ndim == 1:
+            temp = temp + 1
+        else:
+            temp[idx] = temp[idx] + 1
+        assert np.all(padded_mesh.n == temp)
+
+    for dim in mesh.region.dims:
```
fangohr, line 1105:
```suggestion
    for dim in mesh.region.dims:  # adding two layers on pmin side and 3 on pmax, for each dimension
```

fangohr (commented):
First set of feedback.

fangohr:
The constructor does not check if the cell size exceeds the actual region (relevant code ends at https://github.com/ubermag/discretisedfield/pull/255/files#diff-c61c4320b3a6f255741f90a4610a5b2aae26f9698465b9fabab7e35b43b3938eL218)

Here is an example:

```python
In [57]: mesh = df.Mesh(region=df.Region(p1=0, p2=1e-3), cell=1)

In [58]: mesh
Out[58]: Mesh(Region(pmin=[0.0], pmax=[0.001], dims=['x'], units=['m']), n=[0])

In [59]: mesh.dx
/Users/fangohr/Documents/GitHub/discretisedfield/discretisedfield/mesh.py:294: RuntimeWarning: divide by zero encountered in divide
  return np.divide(self.region.edges, self.n).astype(float)
Out[59]: inf
```

We do check if the cell size is too large:
```python
In [60]: mesh = df.Mesh(region=df.Region(p1=0, p2=1e-2), cell=1)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[60], line 1
----> 1 mesh = df.Mesh(region=df.Region(p1=0, p2=1e-2), cell=1)

File ~/Documents/GitHub/discretisedfield/discretisedfield/mesh.py:214, in Mesh.__init__(self, region, p1, p2, n, cell, bc, subregions, attributes)
    210     rem = np.remainder(self.region.edges, cell)
    211     if np.logical_and(
    212         np.greater(rem, tol), np.less(rem, np.subtract(cell, tol))
    213     ).any():
--> 214         raise ValueError(
    215             "Region cannot be divided into "
    216             f"discretisation cells of size {cell=}."
    217         )
    218     self._n = np.divide(self.region.edges, cell).round().astype(int)
    220 elif n is not None and cell is None:
    221     # scalar data types for 1d regions

ValueError: Region cannot be divided into discretisation cells of size cell=[1].
```
but that checks seems to fail if the discrepancy between |p2-p1| and the cell size becomes too large.

I could create an issue for this so we can move on with this PR?

fangohr:
Overall this looks clean and very nicely done.

lang-m:
The bug with cell size > region size has been fixed in 5cdc0bc

PR #256, opened 2022-12-12
Select method for `Mesh` class

swapneelap:

PR #269, opened 2022-12-13
Select method for Field

lang-m:
Based on #256

### DIFF
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -2746,7 +2746,9 @@ class Field(_FieldIO):
                 )
             ]
 
-        return self.__class__(mesh, nvdim=self.nvdim, value=array, vdims=self.vdims)
+        return self.__class__(
+            mesh, nvdim=self.nvdim, value=array, vdims=self.vdims, unit=self.unit
+        )
 
     def plane(self, *args, n=None, **kwargs):
         """Extracts field on the plane mesh.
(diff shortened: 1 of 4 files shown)


lang-m:
One test is currently failing. Before this can be addressed #298 has to be merged.

lang-m:
Locally, all test for field/mesh sel pass.

discretisedfield/field.py
```diff
@@ -2599,6 +2599,105 @@ def line(self, p1, p2, n=100):
...
+            dfu.assemble_index(
+                slice(None), self.mesh.region.ndim + 1, {dim_index: sel_index}
+            )
+        ]
+
+        try:
+            mesh = self.mesh.sel(*args, **kwargs)
+        except ValueError as e:
+            if "p1 and p2 must not be empty" not in str(e):
+                raise
```
swapneelap, line 2694: Does this require a `raise e` or simply `raise`? I have never used this, hence I do not know.

swapneelap (approved):
I really like the `_sel_convert_input` method which avoids duplication in both field and mesh select methods.
I have a minute comment, however, it is more of a question than a suggestion.

discretisedfield/field.py
lang-m: This is on purpose `raise` without an exception and re-raises the currently handled exception.

`raise e` would technically also work but then the stack trace looks a bit different (without the `e` line 2694 does not appear in the stack trace).

samjrholt (approved):

PR #271, opened 2022-12-13
Nd field basic tests2

fangohr:
Same as #258, but without the last two (accidental) merges.

PR #273, opened 2022-12-14
Improve nvdim documentation

fangohr:
based on #271, improve `nvdim` related documentation: make clear we mean the NUMBER of dimensions (i.e. an int, and not the names of the dimensions.)

### DIFF
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -41,8 +41,8 @@ class Field(_FieldIO):
 
     nvdim : int
 
-        Dimension of the field's value. For instance, if `dim=3` the field is a
-        three-dimensional vector field and for `dim=1` the field is a scalar
+        Number of Value DIMensions of the field. For instance, if `nvdim=3` the field is a
+        three-dimensional vector field and for `nvdim=1` the field is a scalar
         field.
 
     value : array_like, callable, dict, optional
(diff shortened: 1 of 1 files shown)


discretisedfield/field.py
```diff
@@ -41,8 +41,8 @@ class Field(_FieldIO):
 
     nvdim : int
 
-        Dimension of the field's value. For instance, if `dim=3` the field is a
-        three-dimensional vector field and for `dim=1` the field is a scalar
+        Number of Value DIMensions of the field. For instance, if `nvdim=3` the field is a
```
samjrholt, line 44: Is DIM capitalised on purpose?
Same with Value

discretisedfield/field.py
```diff
@@ -1401,11 +1401,11 @@ def __mul__(self, other):
 
         It can be applied between:
 
-        1. Two fields with equal vector dimentions,
+        1. Two fields with equal vector dimensions ``nvdim``,
```
samjrholt, line 1404: We seem to use a mixture of `value dimensions` and `vector dimensions` when describing `nvdims`.

samjrholt (changes requested):

discretisedfield/field.py
fangohr: Yes - the idea is that the capitalised letters form the argument name (`nvdim`) together.

discretisedfield/field.py
fangohr: Yes, that is a bit unfortunate. We mean value dimensions but in this particular example those values are vector fields, so the 'v' can be misread. I had noticed but didn't see a good solution.

Perhaps we can get rid of the `vector` here (and potentially replace it by `value`): if both fields are scalar, we can also multiply them?

samjrholt (approved):

PR #275, opened 2022-12-15
Extend region notebook

lang-m:


lang-m:
@swapneelap Could you please have a look at this so that we can merge it.

swapneelap (commented):
In region-basics below cell 3: `In our case $p_\\text{min} = p_{1}$ and $p_\\text{max} = p_{2}$, only because of our choice of diagonally opposite points of the cubic region. In general, this is not the case, since we could have chosen any two points for $p_{1}$ and $p_{2}$.\n",`. I find a bit confusing as p_1 and p_2 will have to be diagonally opposite points. They cannot be any two points in the region. It would be best to re-word this.
This is the only comment I have; everything else looks good!

lang-m:
You are right. I think somewhere in the beginning it says that `p1` and `p2` have to be diagonally opposite. Do you want to suggest a more clear explanation?

swapneelap:
One option can be: `In the above example $p_\\text{min} = p_{1}$ and $p_\\text{max} = p_{2}$, however, this is not the case every time. In general one could have chosen any two diagonally opposite points for $p_{1}$ and $p_{2}$ to define the region, for example, ${p}_{1} = (100\text{e}^{-9}, 0, 0)$ and ${p}_{1} = (0, 50\text{e}^{-9}, 20\text{e}^{-9})$ .`

PR #282, opened 2022-12-16
Add valid

samjrholt:


discretisedfield/tests/test_field.py
```diff
@@ -416,6 +416,113 @@ def test_unit(test_field):
...
+            return 0
+
+    f = df.Field(mesh, nvdim=nvdim, value=(1,) * nvdim, norm=norm_func, valid="norm")
+    assert f.valid.shape == (*mesh.n, 1)
+    assert f.valid.dtype == bool
+    for idx in f.mesh.indices:
+        if all(f.mesh.index2point(idx) < 5):
+            assert f.valid[tuple(idx)][0]
+        else:
+            assert not f.valid[tuple(idx)][0]
```
lang-m, line 465: Again, we can see that `mesh.indices` should return tuples. I'll create a PR for this and change it back.

discretisedfield/tests/test_field.py
```diff
@@ -416,6 +416,113 @@ def test_unit(test_field):
...
+            return 5.0
+        else:
+            return 0
+
+    f = df.Field(mesh, nvdim=nvdim, value=(1,) * nvdim, norm=norm_func, valid="norm")
+    assert f.valid.shape == (*mesh.n, 1)
+    assert f.valid.dtype == bool
+    for idx in f.mesh.indices:
+        if all(f.mesh.index2point(idx) < 5):
+            assert f.valid[tuple(idx)][0]
```
lang-m, line 463: The extra `[0]` is a bit confusing at first (as usual). Maybe add a short comment.

discretisedfield/tests/test_field.py
```diff
@@ -416,6 +416,113 @@ def test_unit(test_field):
...
+    [1, 2, 3, 4],
+)
+def test_valid_set_call(ndim, nvdim):
+    mesh = df.Mesh(p1=(0,) * ndim, p2=(10,) * ndim, cell=(1,) * ndim)
+
+    def valid_func(point):
+        if all(point < 5):
+            return True
+        else:
+            return False
```
lang-m, line 480:
```suggestion
        return all(point < 5)
```

lang-m (commented):
Looks good so far.

A few general questions (also for @swapneelap and @fangohr):
- Did we make a decision for keeping/"forwarding" the `valid` array during operations?
- Do we want to save the valid array (in hdf5)?

fangohr:
> A few general questions (also for @swapneelap and @fangohr):
> 
> * Did we make a decision for keeping/"forwarding" the `valid` array during operations?
You mean when adding two fields, for example? Not sure we did, but it would be useful. For element-wise operations, I would expect that the element in the resulting field is valid only if the elements in both operand fields are valid.

> * Do we want to save the valid array (in hdf5)?
We should I think.

discretisedfield/field.py
```diff
@@ -1075,8 +1118,10 @@ def _check_same_mesh_and_field_dim(self, other, ignore_scalar=False):
             )
 
     def _apply_operator(self, other, function, operator):
+        valid = self.valid.copy()
```
lang-m, line 1121: We do not need a copy here because a copy is automatically created during initialisation of the new field.

```suggestion
        valid = self.valid
```

discretisedfield/field.py
```diff
@@ -1559,8 +1606,10 @@ def dot(self, other):
         array([5.])
 
         """
+        valid = self.valid.copy()
```
lang-m, line 1609:
```suggestion
        valid = self.valid
```

discretisedfield/field.py
```diff
@@ -1706,10 +1760,12 @@ def __lshift__(self, other):
         True
 
         """
+        valid = self.valid.copy()
```
lang-m, line 1763:
```suggestion
        valid = self.valid
```

discretisedfield/field.py
```diff
@@ -1621,6 +1672,7 @@ def cross(self, other):
         array([ 0., -1.,  0.])
 
         """
+        valid = self.valid.copy()
```
lang-m, line 1675:
```suggestion
        valid = self.valid
```

discretisedfield/field.py
```diff
@@ -2665,12 +2725,14 @@ def plane(self, *args, n=None, **kwargs):
...
             value = self.array[slices]
+            valid = self.valid[slices]
         return self.__class__(
             plane_mesh,
             nvdim=self.nvdim,
             value=value,
             vdims=self.vdims,
             unit=self.unit,
+            valid=valid,
         )
```
lang-m, line 2736: I would ignore the plane method (as it will be removed soon).

discretisedfield/field.py
```diff
@@ -2734,6 +2796,7 @@ def resample(self, n):
             vdims=self.vdims,
             unit=self.unit,
             dtype=self.dtype,
+            valid=self.__class__(self.mesh, nvdim=1, value=self.valid),
```
lang-m, line 2799: Do we need `dtype=bool` as additional argument?

discretisedfield/field.py
```diff
@@ -2872,8 +2936,10 @@ def angle(self, vector):
         array([1.57079633])
 
         """
+        valid = self.valid.copy()
```
lang-m, line 2939:
```suggestion
        valid = self.valid
```

lang-m (commented):
One additional more general question: could we benefit from `valid` being a `Field` instead of a numpy array? Or are there any obvious disadvantages?

samjrholt:
> One additional more general question: could we benefit from `valid` being a `Field` instead of a numpy array? Or are there any obvious disadvantages?

This was actually on my list of things to talk to you about! I think it would be beneficial if it was a field. However, I was slightly worried about recursion as our valid field would have a valid property which is a field etc. Thoughts?

fangohr:
> 

What would be the advantages?

samjrholt:
> > 
> 
> What would be the advantages?

The main benefit I can see is we would be able to use the plotting functionality.

fangohr:
> > What would be the advantages?
> 
> The main benefit I can see is we would be able to use the plotting functionality. (and for some internal things like resampling).

Nice idea. Could I create a new field with the `field.valid` array as data to initialise and do the same? 

[Experience has (painfully) shown that often it is good to keep something simple. Which would be in favour of making `valid` an array, not a Field. So I am trying to find a solution for that *if possible*.]

samjrholt:
We can create a valid `df.field` object using of a `field` by
```
df.Field(field.mesh, nvdim=1, value=field.valid, dtype=bool)
```
(I'm not 100% sure if we explicitly need the `dtype`)

Without a major change to the `valid` property I'm not sure how we could make `field.valid` a `df.Field` with it being recursive.

lang-m:
Good point, the `valid` Field would again need a `valid` Field. So we would run into recursions as you pointed out.

Keeping it an array is probably the best solution.

discretisedfield/field.py
```diff
@@ -149,8 +150,10 @@ def __init__(
 
         self.unit = unit
 
+        self.valid = True  # required in here for correct initialisation
```
lang-m, line 153: Can you extend the comment to explain why it is required?

lang-m (approved):

lang-m (approved):

PR #287, opened 2022-12-20
__slots__ in Region, Mesh and Field

lang-m:
A few other small fixes are also in this PR as I noticed these problems while introducing `__slots__` (although they are independent of the slots).

### DIFF
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -688,7 +688,7 @@ class Field(_FieldIO):
     @property
     def average(self):
         raise AttributeError(
-            "This property has been remove, please use the mean method."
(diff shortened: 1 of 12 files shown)


discretisedfield/field.py
```diff
@@ -685,7 +687,9 @@ def mean(self, direction=None):
 
     @property
     def average(self):
-        raise ValueError("This property has been remove, please use the mean method.")
+        raise AttributeError(
+            "This property has been remove, please use the mean method."
```
fangohr, line 691:
```suggestion
            "This property has been removed, please use the mean method."
```

fangohr (approved):
Looks good to me. One minor suggestion.

PR #289, opened 2023-01-09
Nd integrate

lang-m:


lang-m:
Needs #269

discretisedfield/tests/test_field.py
```diff
@@ -1935,14 +1945,36 @@ def value_fun(point):
...
+
+    # 2d mesh with one cell in integration direction
+    mesh = df.Mesh(p1=(0, 0), p2=(10, 1), cell=(1, 1))
+    f = df.Field(mesh, nvdim=2, value=(2, 2.5))
+    f_int = f.integrate("y", cumulative=True)
+    assert f_int.mesh == f.mesh
+    assert np.allclose(f_int.mean(), (1, 1.25))
+
+
+def test_integrate_exceptions():
```
samjrholt, line 1964: Looks good, shall we also check for a list of directions as this may be a common error? I realise you are already testing for an `int` type.

samjrholt (approved):

discretisedfield/tests/test_field.py
lang-m: Good idea. Could you please suggest a test for this.

PR #298, opened 2023-01-10
mesh getitem

samjrholt:


discretisedfield/mesh.py
```diff
@@ -781,7 +781,7 @@ def point2index(self, point, /):
         if point not in self.region:
             raise ValueError(f"Point {point} is outside the region {self.region=}.")
 
-        index = ((point - self.region.pmin) / self.cell - 0.5).round().astype(int)
+        index = np.floor((point - self.region.pmin) / self.cell).astype(int)
```
lang-m, line 784: In the documentation we should mention that cells (apart from the last one) are half-open intervals.

discretisedfield/mesh.py
```diff
@@ -1363,7 +1363,10 @@ def __getitem__(self, item):
 
         hc = np.divide(self.cell, 2)  # half-cell
         p1 = np.subtract(self.index2point(self.point2index(item.pmin)), hc)
-        p2 = np.add(self.index2point(self.point2index(item.pmax)), hc)
+
+        # Calculate p2 index manually as point2index will give [) and we want [].
```
lang-m, line 1367: Mention in the documentation that regions -> __getitem__ is a closed interval.

discretisedfield/tests/test_mesh.py
```diff
@@ -703,6 +703,46 @@ def test_point2index_valid(p1, p2, n, point, expected):
...
+    assert mesh.point2index(point) == 9
+
+    # Check out of bounds
+    point = -10
+    with pytest.raises(ValueError):
+        mesh.point2index(point)
+
+    point = 20
+    with pytest.raises(ValueError):
+        mesh.point2index(point)
```
lang-m, line 743: You could consider removing the `point` variables, moving the values into the method calls and delete some of the empty lines to make it more concise and group blocks of code around individual boundaries.

lang-m (approved):
Just a few small comments.

PR #300, opened 2023-01-11
Update topological charge

lang-m:
Required changes:
- operate on 2d meshes
- remove mesh.attributes
- fix diff problems (@samjrholt )

Based on #289


lang-m:
Wait for #306

discretisedfield/tests/test_tools.py
```diff
@@ -17,51 +18,45 @@ def test_topological_charge():
...
+        with pytest.raises(ValueError):
+            getattr(dft, function)(f.sel("z"), method=method)
 
     # Method does not exist
     f = df.Field(mesh, nvdim=3, value=(1, 2, 3))
     for function in ["topological_charge_density", "topological_charge"]:
         with pytest.raises(ValueError):
-            getattr(dft, function)(f.plane("z"), method="wrong")
+            getattr(dft, function)(f.sel("z"), method="wrong")
+            getattr(dft, function)(f.sel("z"), method="wrong")
```
lang-m, line 59: This seems to be duplicated.

discretisedfield/tools/tools.py
```diff
@@ -105,52 +105,59 @@ def topological_charge_density(field, /, method="continuous"):
...
         raise ValueError(msg)
 
-    axis1 = field.mesh.attributes["axis1"]
-    axis2 = field.mesh.attributes["axis2"]
     of = field.orientation  # unit field - orientation field
 
+    # Spatial axes associated with the first and second vector
+    # dimensions.
+    axis1 = field.vdim_mapping[field.vdims[0]]
+    axis2 = field.vdim_mapping[field.vdims[1]]
```
lang-m, line 123: We should consider adding some checks here as `vdim_mapping` may be empty or return axes that are not part of `region.dims`.

discretisedfield/tools/tools.py
```diff
@@ -290,18 +299,24 @@ def topological_charge(field, /, method="continuous", absolute=False):
             return float(q.integrate())
 
     elif method == "berg-luescher":
-        axis1 = field.mesh.attributes["axis1"]
-        axis2 = field.mesh.attributes["axis2"]
+        # Spatial axes associated with the first and second vector
+        # dimensions.
+        axis1 = field.vdim_mapping[field.vdims[0]]
+        axis2 = field.vdim_mapping[field.vdims[1]]
```
lang-m, line 305: Again, `axis1` and `axis2` are not guaranteed to be valid and we should maybe add an explicit check.

discretisedfield/tools/tools.py
```diff
@@ -105,52 +105,59 @@ def topological_charge_density(field, /, method="continuous"):
...
 
-        for i, j in itertools.product(range(of.mesh.n[axis1]), range(of.mesh.n[axis2])):
-            index = dfu.assemble_index(0, 3, {axis1: i, axis2: j})
+        for i, j in itertools.product(
+            range(of.mesh.n[axis1_idx]), range(of.mesh.n[axis2_idx])
+        ):
+            index = dfu.assemble_index(0, 2, {axis1_idx: i, axis2_idx: j})
             v0 = of.array[index]
 
             # Extract 4 neighbouring vectors (if they exist)
```
lang-m, line 143: We should probably also use valid in here? What do you think @samjrholt. If yes, I would prefer a separate PR for this.

lang-m (commented):
Due to changed variable names quite a few lines are now too long. We could consider:
`axisX` -> `dimX` and `axisX_idx` -> `axisX` (what we had before).

I left a few comments for additional checks. I am not sure if we really need those. We should have a look at the error messages without and then decide if they are too cryptic.

discretisedfield/tools/tools.py
samjrholt: If I move `axis1_idx = field.mesh.region._dim2index(axis1)` directly below then this should fix this as it has inbuilt error checking. Would you be happy with this?

discretisedfield/tools/tools.py
samjrholt: Yes you are correct, I have created #313.

discretisedfield/tools/tools.py
lang-m: Yes, I think that should be sufficient.

lang-m:
I am happy for this to be merged when the error-checking is done.

PR #303, opened 2023-01-11
Rename `comp` to `vdims` in field xarray methods

swapneelap:


discretisedfield/field.py
```diff
@@ -3317,7 +3317,7 @@ def to_xarray(self, name="field", unit=None):
         """Field value as ``xarray.DataArray``.
 
         The function returns an ``xarray.DataArray`` with dimensions ``x``,
-        ``y``, ``z``, and ``comp`` (only if ``field.nvdim > 1``). The coordinates
+        ``y``, ``z``, and ``vdims`` (only if ``field.nvdim > 1``). The coordinates
```
lang-m, line 3320: We should also generalise the spatial components.

discretisedfield/field.py
```diff
@@ -3430,11 +3430,11 @@ def from_xarray(cls, xa):
         The class method accepts an ``xarray.DataArray`` as an argument to
         return a ``discretisedfield.Field`` object. The DataArray must have
         either three (``x``, ``y``, and ``z`` for a scalar field) or four
```
lang-m, line 3432: This is also outdated as we can now have any number of spatial dimensions.

lang-m (approved):

swapneelap:
@lang-m you are right about the docstrings. I was thinking of doing this in a separate PR dedicated to updating the docstrings. What do you think?

lang-m:
You are right, we have a lot of outdated docstrings. So a separate PR sounds good.

PR #304, opened 2023-01-11
Update mpl plotting

lang-m:
- [x] Better checks for vdim-dim-mapping for `vector`
- [x] Lightness plots do not work because `angle` returns a vector between 0 and π (instead of 2π). @samjrholt Do we change `angle` or change the implementation for `lightness`? Edit: `lightness can be re-implemented based on #309 

Based on #302


lang-m:
@samjrholt Please ignore the changes in the notebook.

samjrholt (approved):
Looks good. Is it worth using the `valid` array as the default `filter_field`?

PR #307, opened 2023-01-11
Change shape of valid

samjrholt:
The shape of valid should be `field.valid.shape = field.mesh.n` not `field.valid.shape = (*field.mesh.n, 1)`. This will allow proper Boolean indexing e.g. `field.array[field.valid]`.

### DIFF
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -521,8 +521,10 @@ class Field(_FieldIO):
                 valid = ~np.isclose(self.norm.array, 0)
         else:
             valid = True
-
-        self._valid = _as_array(valid, self.mesh, nvdim=1, dtype=bool)
+        # Using _as_array creates an array with shape (*mesh.n, 1).
+        # We only want a shape of mesh.n so we can directly use it
+        # to index field.array i.e. field.array[field.valid].
+        self._valid = _as_array(valid, self.mesh, nvdim=1, dtype=bool)[..., 0]
 
     def __abs__(self):
         """Absolute value of the field.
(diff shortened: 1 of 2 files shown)


discretisedfield/field.py
```diff
@@ -3779,13 +3776,19 @@ def _(val, mesh, nvdim, dtype):
...
         raise ValueError(
             f"Wrong dimension {len(val)} provided for value; expected dimension is"
             f" {nvdim}"
         )
     dtype = dtype or max(np.asarray(val).dtype, np.float64)
-    return np.full((*mesh.n, nvdim), val, dtype=dtype)
+    if np.array_equal(np.shape(val), mesh.n):
+        return np.expand_dims(val, axis=-1)
+    else:
+        return np.full((*mesh.n, nvdim), val, dtype=dtype)
```
lang-m, line 3791:
```suggestion
    if isinstance(val, collections.abc.Iterable):
        if nvdim == 1 and np.array_equal(np.shape(val), mesh.n):
            return np.expand_dims(val, axis=-1)
        elif np.shape(val)[-1] != nvdim:
            raise ValueError(
                f"Wrong dimension {len(val)} provided for value; expected dimension is"
                f" {nvdim}."
            )
    dtype = dtype or max(np.asarray(val).dtype, np.float64)
    return np.full((*mesh.n, nvdim), val, dtype=dtype)
```

lang-m (changes requested):
I tried to simplify the logic in `_as_array` and make it more readable.

Everything else looks good.

PR #323, opened 2023-01-13
318 boundary conditions in diff

samjrholt:


discretisedfield/field.py
```diff
@@ -2358,46 +2308,54 @@ def diff(self, direction, order=1, restrict2valid=True):
...
-                    self.mesh.cell[direction_idx],
-                    bc,
+                    field.mesh.cell[direction_idx],
                 )
 
+        # Remove the padding if periodic is True
+        if periodic:
+            slices = dfu.assemble_index(
+                slice(None), field.mesh.region.ndim + 1, {direction_idx: slice(1, -1)}
+            )
```
lang-m, line 2349: I think we could also use `self.mesh.region2slices`. Might be a little more obvious. (Please test first!)

discretisedfield/field.py
```diff
@@ -2358,46 +2308,54 @@ def diff(self, direction, order=1, restrict2valid=True):
...
+            out = out[slices]
+
         return self.__class__(
             self.mesh,
             nvdim=self.nvdim,
             value=out,
             vdims=self.vdims,
             unit=self.unit,
-            valid=valid,
+            valid=self.valid,
```
lang-m, line 2358: We should in here save the `vdim_mapping`.

discretisedfield/tests/test_field.py
```diff
@@ -1731,15 +1731,15 @@ def value_fun(point):
 
     # No PBC
     f = df.Field(mesh_nopbc, nvdim=1, value=value_fun)
-    assert np.allclose(f.diff("x")((11, 1, 1)), 1)
-    assert np.allclose(f.diff("y")((1, 7, 1)), 1)
-    assert np.allclose(f.diff("z")((1, 1, 5)), 1)
+    assert np.allclose(f.diff("x", periodic=False)((11, 1, 1)), 1)
+    assert np.allclose(f.diff("y", periodic=False)((1, 7, 1)), 1)
+    assert np.allclose(f.diff("z", periodic=False)((1, 1, 5)), 1)
```
lang-m, line 1736: We should probably omit `periodic=False` to test the default and notice accidental changes.

lang-m (changes requested):

lang-m (approved):

PR #327, opened 2023-01-16
Mesh allclose

### LINKED ISSUE #284: Refactor `mesh.allclose` method
At present the use of `atol` and `rtol` in the method is a bit confusing. Moreover, the method fails if one of the points in the two meshes that are equal to or close to 0.0. We can refactor it to take into account the exponent of the cell dimensions, for example.

lang-m:
closes #284

### DIFF
--- a/discretisedfield/tests/test_mesh.py
+++ b/discretisedfield/tests/test_mesh.py
@@ -494,7 +494,6 @@ def test_eq(p1_1, p1_2, p2, n1, n2):
     assert not mesh2 == "mesh2"
 
 
-@pytest.mark.xfail()
 @pytest.mark.parametrize(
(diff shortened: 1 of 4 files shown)


discretisedfield/tests/test_mesh.py
```diff
@@ -551,6 +547,29 @@ def test_allclose(p1_1, p1_2, p2, n1, n2):
         mesh1.allclose(mesh3, rtol="1")
 
 
+def test_allclose_cell_accuracy():
```
lang-m, line 550: @fangohr This test outlines potential problems with the default `region.tolerance_factor` in `mesh.allclose`. However, it is probably not practically relevant.

samjrholt (approved):

PR #337, opened 2023-01-31
Fix ovf extend scalar

lang-m:
OVF files saved with the option `extend_scalar=True` (saving a scalar field as the x component of a vector field) only contain a part of the data when the array size exceeds 100_000 elements (the chunk size).

This PR fixes the bug and updates the test that should detect it (which it now does).


lang-m:
Added to the changelog in commit https://github.com/ubermag/ubermag.github.io/pull/21/commits/fc276f68679bb76538b1dde4db7e17732a0b94b1

PR #340, opened 2023-02-09
Fix: calculation of the topological charge

### LINKED ISSUE #339: Bug in computing topological charge
When trying to compute the topological charge density in a plane (y,z) using the following code:
```python
import discretisedfield as df
import discretisedfield.tools as dft
p1 = (0, 0, 0)
p2 = (10, 10, 10)
cell = (2, 2, 2)
mesh = df.Mesh(p1=p1, p2=p2, cell=cell)
f = df.Field(mesh, nvdim=3, value=(1, 1, 1))
dft.topological_charge_density(f.sel('x'))
```
we get the following error
```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[26], line 8
      6 mesh = df.Mesh(p1=p1, p2=p2, cell=cell)
      7 f = df.Field(mesh, nvdim=3, value=(1, 1, 1))
----> 8 dft.topological_charge_density(f.sel('x'))
      9 #dft.topological_charge_density(f.sel('z'), method='berg-luescher')

File ~/ubermag_devtools/repos/discretisedfield/discretisedfield/tools/tools.py:124, in topological_charge_density(field, method)
    122 axis1 = field.vdim_mapping[field.vdims[0]]
    123 axis2 = field.vdim_mapping[field.vdims[1]]
--> 124 axis1_idx = field.mesh.region._dim2index(axis1)
    125 axis2_idx = field.mesh.region._dim2index(axis2)
    127 if method == "continuous":

File ~/ubermag_devtools/repos/discretisedfield/discretisedfield/region.py:284, in Region._dim2index(self, dim)
    282     return self.dims.index(dim)
    283 except ValueError:
--> 284     raise ValueError(f"'{dim}' not in region.dims={self.dims}.") from None

ValueError: 'x' not in region.dims=('y', 'z').
```

lang-m:
The calculation of the topological charge (density) did mix spatial dimensions and vector dimensions. However, derivatives and integration should only depend on the spatial dimensions (and axis names) of the mesh/region. The bug was introduced during refactoring due to confusion about `axisdict` and `raxesdict`.

closes #339

### DIFF
--- a/discretisedfield/tools/tools.py
+++ b/discretisedfield/tools/tools.py
@@ -102,18 +102,15 @@ def topological_charge_density(field, /, method="continuous"):
 
     """
     if field.nvdim != 3:
-        msg = f"Cannot compute topological charge density for {field.nvdim=} field."
-        raise ValueError(msg)
+        raise ValueError(
+            f"Cannot compute topological charge density for {field.nvdim=} field."
+        )
 
     if field.mesh.region.ndim != 2:
         raise ValueError(
             "The topological charge density can only be computed on fields with 2"
             f" spatial dimensions, not {field.mesh.region.ndim=}."
         )
-
-    if method not in ["continuous", "berg-luescher"]:
-        msg = "Method can be either continuous or berg-luescher"
-        raise ValueError(msg)
 
     of = field.orientation  # unit field - orientation field
 
@@ -123,7 +120,6 @@ def topological_charge_density(field, /, method="continuous"):
         return 1 / (4 * np.pi) * of.dot(of.diff(axis1).cross(of.diff(axis2)))
 
     elif method == "berg-luescher":
-
         q = df.Field(field.mesh, nvdim=1)
 
         # Area of a single triangle
@@ -159,11 +155,15 @@ def topological_charge_density(field, /, method="continuous"):
 
             if triangle_count > 0:
                 q.array[i, j] = charge / (area * triangle_count)
-            else:
-                # If the cell has no neighbouring cells
+            else:  # the cell has no neighbouring cells
                 q.array[i, j] = 0
 
         return q
+
+    else:
+        raise ValueError(
+            f"'method' can be either 'continuous' or 'berg-luescher', not '{method}'."
+        )
 
 
 def topological_charge(field, /, method="continuous", absolute=False):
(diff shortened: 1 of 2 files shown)


discretisedfield/tools/tools.py
```diff
@@ -117,12 +117,9 @@ def topological_charge_density(field, /, method="continuous"):
...
 
-    # Spatial axes associated with the first and second vector
-    # dimensions.
-    axis1 = field.vdim_mapping[field.vdims[0]]
-    axis2 = field.vdim_mapping[field.vdims[1]]
-    axis1_idx = field.mesh.region._dim2index(axis1)
-    axis2_idx = field.mesh.region._dim2index(axis2)
+    axis1 = field.mesh.region.dims[0]
+    axis2 = field.mesh.region.dims[1]
+    axis1_idx, axis2_idx = 0, 1
```
swapneelap, line 122: It looks like `axis1_idx` and `axis2_idx` will remain 0 and 1, respectively, regardless of the field used for calculation of the topological charge density. Maybe we can move this over line 120 and use:
```Python
axis1 = field.mesh.region.dims[axis1_idx]
axis2 = field.mesh.region.dims[axis2_idx]
``` 
Do correct me if I understand this wrong.

swapneelap (commented):
I have a minor comment. One thing: I think we can drop the use of `dfu.assemble_index` as it making the code confusing (at least to me). We can also drop the `axis1_idx` and `axis2_idx` altogether as their values are 0 and 1 regardless (correct me if I am missing something).

discretisedfield/tools/tools.py
```diff
@@ -117,44 +117,28 @@ def topological_charge_density(field, /, method="continuous"):
...
 
     elif method == "berg-luescher":
 
         q = df.Field(field.mesh, nvdim=1)
 
         # Area of a single triangle
-        area = 0.5 * field.mesh.cell[axis1_idx] * field.mesh.cell[axis2_idx]
+        area = 0.5 * field.mesh.cell[0] * field.mesh.cell[1]
+
+        n0, n1 = of.mesh.n[0], of.mesh.n[1]
```
swapneelap, line 132: If you want to reduce the variables, you can directly use `of.mesh.n[0]` and `of.mesh.n[1]`below.

swapneelap (approved):
Left a minor comment. Code looks more readable. The Berg-Luescher method logic looks good to me.

PR #371, opened 2023-04-25
Fix test for mesh.allclose

lang-m:
limited data range for integers on Windows


lang-m:
The failing unit test does pass now. Overall, tests cannot pass (yet) because of the outdated docstrings and notebooks.

fangohr (approved):

PR #373, opened 2023-06-23
Fix: valid initialisation

lang-m:
When setting valid using a numpy array the comparisons with 'norm' would be done element-wise.


swapneelap (approved):
Looks good to me!

PR #374, opened 2023-06-23
Fix: mesh comparison for numerical operations

lang-m:
The checks for the mesh did not take limited numerical precision into account. I think this is a left-over from the refactoring.


swapneelap:
Looks good to merge from my side.

PR #376, opened 2023-07-10
Fix failing tests as on 10/07/2023

swapneelap:


swapneelap:
This is a `matplotlib` problem which is fixed upstream and should go away in version `3.7.3`. Please have a look at [this](https://github.com/matplotlib/matplotlib/issues/26287) issue. Should we close this PR @lang-m?

lang-m:
I would not just remove `bbox_inches="tight"` without further testing because I am not sure what side effects that may have in the generated files.
We could
- either: do some tests with different geometries to see if the option is useful in general
- or you check with the last few versions of matplotlib where it fails (maybe the newest 3.6.X and 3.7.[0,1,2]). Then we could exclude the problematic versions from the list of supported matplotlib versions in our requirements.

swapneelap:
@lang-m unit tests are passing now. If you can have a look and approve the PR, I'll merge it.

PR #377, opened 2023-07-13
Enhancements to HDF5 Saving and Loading Mechanism

lang-m:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces enhancements to the HDF5 saving and loading mechanism in the discretisedfield library. It adds the ability to save and load the validity information of the data as a separate dataset in HDF5 files. It also includes a check to save subregions only if they are specified. Additionally, it fixes an issue with loading boundary conditions from HDF5 files.

___
## PR Main Files Walkthrough:
`discretisedfield/io/hdf5.py`: The changes in this file include the addition of a new method `_h5_save_valid` to save the validity information of the data. The `_h5_save` and `_h5_load` methods have been updated to handle subregions only if they are specified. The `_h5_load` method has also been updated to load boundary conditions from HDF5 files. The `_h5_load_field` method now includes the loading of validity information.


samjrholt:
@samjrholt @swapneelap Add tests

lang-m:
/ask We have one attribute `ubermag-hdf5-file-version: '0.1'`. Should we use `'0.1'` or `'v0.1'`?

samjrholt (approved):

samjrholt:
Consider adding comments regarding not saving subregions into the file rather than it being empty.

lang-m:
> Consider adding comments regarding not saving subregions into the file rather than it being empty.

Do you mean in the docstring or in the code [for developers]?

The older version of this implementation with empty subregions in the HDF5 file was never released (only on `master` for a few months) so I think "rather than being empty" would not make. I will in any case extend the documentation to say more specifically that the subregion information is only present if the field contains subregions.

Let me know, if you think I should also add a comment inside the source code.

samjrholt:
> > Consider adding comments regarding not saving subregions into the file rather than it being empty.
> 
> Do you mean in the docstring or in the code [for developers]?
> 
> The older version of this implementation with empty subregions in the HDF5 file was never released (only on `master` for a few months) so I think "rather than being empty" would not make. I will in any case extend the documentation to say more specifically that the subregion information is only present if the field contains subregions.
> 
> Let me know, if you think I should also add a comment inside the source code.

Just the documentation

PR #378, opened 2023-07-14
Bug: wrong axis labels in k3d

lang-m:
String formatting was not correct, so all axis had the label `dim nm` instead of e.g. `x (nm)`.


swapneelap (approved):
Looks good to me.

PR #380, opened 2023-07-27
Review and update `Region` docstrings

swapneelap:


discretisedfield/region.py
```diff
@@ -136,9 +143,9 @@ def pmin(self):
 
         Returns
         -------
-        numpy.ndarray (3,)
+        numpy.ndarray
 
-            Point with minimum coordinates :math:`(p_x^\text{min},
+            Point with minimum coordinates for tree dimensions :math:`(p_x^\text{min},
```
samjrholt, line 148:
```suggestion
            Point with minimum coordinates. E.g. for three dimensions :math:`(p_x^\text{min},
```

samjrholt: @swapneelap does this not give an array of `len(ndims)`?

discretisedfield/region.py
```diff
@@ -364,7 +371,7 @@ def edges(self):
 
         Returns
         -------
-        numpy.ndarray (3,)
+        numpy.ndarray
 
              Edge lengths :math:`(l_{x}, l_{y}, l_{z})`.
```
samjrholt, line 376:
```suggestion
             Edge lengths 3 dimensions :math:`(l_{x}, l_{y}, l_{z})`.
             Edge lengths n dimensions :math:`(l_{0}, ..., l_{n})`.
```

What do you think of something like this?
We could add the default n dimension names.

If we wanted something like this I could go through and edit all the doc strings with this example

swapneelap: I like the the `n` dimensions idea. If you like, please go ahead and edit it everywhere.

discretisedfield/region.py
swapneelap: @samjrholt Sorry I did not get your point.

PR #382, opened 2023-07-28
Update and review `Field` docstrings

swapneelap:


discretisedfield/field.py
```diff
@@ -2304,8 +2322,8 @@ def diff(self, direction, order=1, restrict2valid=True, periodic_bc=False):
         ...
         >>> mesh = df.Mesh(p1=0, p2=10, cell=1)
         >>> f = df.Field(mesh, nvdim=1, value=value_fun, valid=valid_fun)
-        >>> f.diff('x', order=1)
-        array([1., 1., 1., 1., 1., 0., 0., 0., 0., 0.])
+        >>> f.diff('x', order=1).array.tolist()
```
samjrholt, line 2325: Why do we need the `tolist()`?

samjrholt (approved):

discretisedfield/field.py
swapneelap: I did this because the output of `f.diff('x', order=1)` comes out to be something like:
```python
array([1.,
       1.,
       1.,
       1.,
       1.,
       0.,
       0.,
       0.,
       0.,
       0.])
```
I was thinking it would add unnecessary lines to the docstrings but `array.tolist()` also looks ugly. I am not sure here, let me know if you have a preference.

PR #384, opened 2023-07-28
Update and review discretisedfield tools docstrings

swapneelap:
I am not sure if the the functions of `tools` actually support the new `discretizedfield` functionality but I think it is good to make the docstrings generic to incorporate the n dimensional fields.

### DIFF
--- a/discretisedfield/tools/tools.py
+++ b/discretisedfield/tools/tools.py
@@ -11,9 +11,9 @@ import discretisedfield.util as dfu
 def topological_charge_density(field, /, method="continuous"):
     r"""Topological charge density.
 
-    This method computes the topological charge density for a vector field
-    (``nvdim=3``). Two different methods are available and can be selected using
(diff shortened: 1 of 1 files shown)


discretisedfield/tools/tools.py
```diff
@@ -11,9 +11,9 @@
 def topological_charge_density(field, /, method="continuous"):
     r"""Topological charge density.
 
-    This method computes the topological charge density for a vector field
-    (``nvdim=3``). Two different methods are available and can be selected using
-    ``method``:
+    This method computes the topological charge density for a vector field having three
+    components (i.e. ``nvdim=3``). Two different methods are available and can be
```
samjrholt, line 15:
```suggestion
    value components (i.e. ``nvdim=3``). Two different methods are available and can be
```

samjrholt (approved):

PR #394, opened 2023-08-02
update notebook field rotations

kzqureshi:


swapneelap:
@samjrholt can you please have a look at the contents of this notebook. I think you have a better idea about rotation than me.

PS: the notebook tests are passing.

samjrholt:
@swapneelap I've removed `k3d` as we were planning to do this in all notebooks and have also added a statement about 3D meshes.

samjrholt (approved):

PR #409, opened 2023-08-03
fft notebook

samjrholt:
Dependent on:

- [x] #410 

### Summary by CodeRabbit

- New Feature: Improved performance of FFT calculations by utilizing the faster `scipy` library and optimizing computation.
- Documentation: Added a comment to explain the purpose of a code block in `discretisedfield/mesh.py`.
- Test: Modified and added assertions for FFT calculations in `discretisedfield/tests/test_field.py`, ensuring correctness in various scenarios.
- Test: Added additional tests for FFT calculations, including testing 1D rfft and rfftn with no shift in the last dimension.
- Test: Created a new field with specific values and compared its FFT with the expected result using `scipy.fft.rfftn`.


samjrholt:
@lang-m When you get a chance could you please look through this. There may be a bug in `rfft` as the peak in the spectrum is not in the same place.

So the current example is probably more useful for people doing micromagnetics but I also considered changing the example to 2 spatial dimensions and a scalar field as this simplifies the code/plots a little.

samjrholt:
I have fixed the bug in #410. Please review that first and then merge into this one then please review this.

docs/field-fft.ipynb
```diff
@@ -0,0 +1,668 @@
...
+   "source": [
+    "import discretisedfield as df"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5c652ca9-be64-4cef-8a73-0fa617a668ec",
+   "metadata": {},
+   "source": [
+    "As an example we study a cuboidal geometry with a helix propagating along the `z` direction with a $7, nm$ wavelength."
```
lang-m, line 34:
```suggestion
    "As an example we study a cuboidal geometry with a helix propagating along the `z` direction with a $7\,\mathrm{nm}$ wavelength."
```

docs/field-fft.ipynb
```diff
@@ -0,0 +1,668 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "8d2f8678",
+   "metadata": {},
+   "source": [
+    "### FFT Methods\n",
+    "\n",
+    "The `discretisedfield` package provides four different methods for performing Fast Fourier Transform (FFT) operations on a `discretisedfield.Field` object:\n",
+    "\n",
+    "1. `fftn`: This is the \"full\" FFT method which computes the n-dimensional FFT. It returns the complex-valued FFT result that has the same shape as the input field. It's used when the full frequency spectrum is needed.\n",
```
lang-m, line 124: "Frequency" sounds a bit like we would start from time, not space. Can we make more clear that it is "frequencies" in k space?

docs/field-fft.ipynb
```diff
@@ -0,0 +1,668 @@
...
+    "\n",
+    "3. `rfftn`: This stands for \"real FFT\". It's similar to `fftn`, but it only computes the non-negative frequency terms. The result is a complex-valued array that has half the size of the last dimension compared to the input. It's used when the input field only has real values, and hence its frequency spectrum is symmetric. Thus, it saves computation time and storage.\n",
+    "\n",
+    "4. `irfftn`: This is the inverse of the \"real FFT\" method. It takes a half-sized complex-valued FFT result (as output by `rfftn`) and computes the inverse FFT to return a real-valued field in the spatial domain. It's used when the original field is known to be real-valued.\n",
+    "\n",
+    "All of these methods are `n` dimensional and work with any number of value (vector) and mesh(geometric) dimensions. These four methods provide the necessary tools to perform FFT operations and allow the transition between the spatial domain and frequency domain, covering both real and complex-valued fields.\n",
```
lang-m, line 132:
```suggestion
    "All of these methods are `n` dimensional and work with any number of value (vector) and mesh (geometric) dimensions. These four methods provide the necessary tools to perform FFT operations and allow the transition between the spatial domain and frequency domain, covering both real and complex-valued fields.\n",
```

docs/field-fft.ipynb
```diff
@@ -0,0 +1,668 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "46494862",
+   "metadata": {},
+   "source": [
+    "With the application of the FFT methods in `discretisedfield`, notable transformations occur in both the `mesh` and the `field`:\n",
+    "\n",
+    "1. **Mesh Transformation**: The `mesh` is transposed from the regular space to the spatial frequency domain. As a consequence, the units transition from meters ($m$) to inverse meters ($(m)^{-1}$). Furthermore, to signify this transformation, `k_` is prepended to each of the `dims` of the mesh. However, the Fourier transformed `mesh` maintains the same number of points (`n`) as the original mesh.\n",
```
lang-m, line 191: Units should not be italic

docs/field-fft.ipynb
```diff
@@ -0,0 +1,668 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "2f867782",
+   "metadata": {},
+   "source": [
+    "This discrepancy stems primarily from two factors:\n",
+    "\n",
+    "1. **Floating-Point Precision**: Arithmetic operations on floating-point numbers can lead to small imprecisions.\n",
+    "\n",
+    "2. **Complex Values**: Initially, our field comprised real values. However, the processing via FFT and inverse FFT results in a field with complex values. \n",
```
lang-m, line 408: Maybe we can add two additional plots to show that the real part is mostly the initial field and the imaginary part practically zero?

lang-m (commented):
The notebook looks really good. Here is some preliminary feedback from a first review. I'll go over it again more thoroughly after reviewing and merging the bug fix.

docs/field-fft.ipynb
```diff
@@ -0,0 +1,1697 @@
...
+   "source": [
+    "ifft_field.imag.mpl()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "72133a4a",
+   "metadata": {},
+   "source": [
+    "As we can see, the real part of the inverse Fourier transformed field is almost identical to the original field, while the imaginary part is practically zero."
```
lang-m, line 446:
```suggestion
    "As we can see, the real part of the inverse Fourier transformed field is visually identical to the original field, while the imaginary part is practically zero."
```

docs/field-fft.ipynb
```diff
@@ -0,0 +1,1697 @@
...
+    "\n",
+    "psd.mpl()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3b4bd018",
+   "metadata": {},
+   "source": [
+    "We can see that the mesh and the plot look different for real Fourier transforms compared to full Fourier transforms. For Fourier transforms, the `mesh` is converted from the standard spatial domain to the spatial frequency domain but for real Fourier transforms there are some key differences. As before, each dimension of the mesh (`dims`) is prepended with `k_` to signify this change. However, the Fourier transformed `mesh` no longer has the the same number of points (`n`) as the original mesh in all directions. In the final dimension, the number of points is reduced by half (If n is even, the length of the transformed axis is $(n/2)+1$. If n is odd, the length is $(n+1)/2$).\n",
```
lang-m, line 535: For better clarity:
```suggestion
    "We can see that the mesh and the plot look different for real Fourier transforms compared to full Fourier transforms. For full Fourier transforms, the `mesh` is converted from the standard spatial domain to the spatial frequency domain but for real Fourier transforms there are some key differences. As before, each dimension of the mesh (`dims`) is prepended with `k_` to signify this change. However, the Fourier transformed `mesh` no longer has the the same number of points (`n`) as the original mesh in all directions. In the final dimension, the number of points is reduced by half (If n is even, the length of the transformed axis is $(n/2)+1$. If n is odd, the length is $(n+1)/2$).\n",
```

docs/field-fft.ipynb
```diff
@@ -0,0 +1,1697 @@
...
+    "\n",
+    "This halving is a result of the properties of real-valued functions, which have a Hermitian-symmetric Fourier transform. As a result, it is sufficient to store only half of the frequency components (those for non-negative frequencies).\n",
+    "\n",
+    "### irfft\n",
```
lang-m, line 539:
```suggestion
    "### irFFT\n",
```

docs/field-fft.ipynb
```diff
@@ -0,0 +1,1697 @@
...
+    "As we have an rFFT field we can convert this back to the spatial domain by using an irfft."
```
lang-m, line 540:
```suggestion
    "As we have an rFFT field we can convert this back to the spatial domain by using an irFFT."
```

lang-m (commented):
Final review for the first part of the notebook (anything before "Visualising and analysing FFTs"). I have added some more suggestions and minor corretions.

I'll try to also review the second half of the notebooks soon.

docs/field-fft.ipynb
```diff
@@ -0,0 +1,1697 @@
...
+    }
+   ],
+   "source": [
+    "import numpy as np\n",
+    "\n",
+    "mesh = df.Mesh(p1=(0, 0), p2=(20, 45), cell=(1, 1))\n",
+    "\n",
+    "\n",
+    "def value(p):\n",
+    "    x, y = p\n",
```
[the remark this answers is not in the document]
samjrholt: @coderabbitai What is your opinion accounting for ease of use and best practices?

lang-m (approved):
Some additional minor comments for the second half of the notebook.

Feel free to merge once all open comments are addressed.

Edit: several of the comments seem to be off by one line (not sure how that has happened). I hope they all make sense. Directly applying the suggestions of course doesn't work for these.

docs/field-fft.ipynb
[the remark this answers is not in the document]
lang-m: I would use the underscore. We can add a comment if you think that would be useful

PR #410, opened 2023-08-03
no fftshift on last dim for real ffts

samjrholt:
For `rfft` the last dimension should not be shifted! This is because all other dimensions are from `-k` to `+k` but the last dimension is from `0` to `+k`

### DIFF
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -3520,7 +3520,7 @@ class Field(_FieldIO):
         axes = range(self.mesh.region.ndim)
         ft = spfft.fftshift(
             spfft.rfftn(self.array, axes=axes, **kwargs),
-            axes=axes,
+            axes=axes[:-1],
         )
 
         return self._fftn(mesh=mesh, array=ft, ifftn=False)
@@ -3566,7 +3566,7 @@ class Field(_FieldIO):
 
         axes = range(self.mesh.region.ndim)
         ft = spfft.irfftn(
-            spfft.ifftshift(self.array, axes=axes),
+            spfft.ifftshift(self.array, axes=axes[:-1]),
             axes=axes,
             s=shape,
             **kwargs,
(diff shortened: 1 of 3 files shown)


samjrholt:
The tests will be updated to use parameterisation in a later pull request

discretisedfield/tests/test_field.py
```diff
@@ -2709,6 +2710,27 @@ def _init_random(p):
 
     assert f.allclose(f.rfftn().irfftn(shape=f.mesh.n))
 
+    # test 1d rfft
+    assert f.allclose(f.rfftn().irfftn(shape=f.mesh.n).real)
+    assert df.Field(mesh, nvdim=1).allclose(f.rfftn().irfftn(shape=f.mesh.n).imag)
+
+    assert f.allclose(f.rfftn().irfftn(shape=f.mesh.n))
```
lang-m, line 2717: Why do we check first `real` and `imag` individually and afterwards the the full field?

discretisedfield/field.py
```diff
@@ -3520,7 +3520,7 @@ def rfftn(self, **kwargs):
         axes = range(self.mesh.region.ndim)
         ft = spfft.fftshift(
             spfft.rfftn(self.array, axes=axes, **kwargs),
-            axes=axes,
+            axes=axes[:-1],
         )
```
lang-m, line 3524: Why is there no `fftshift` in the mesh? https://github.com/ubermag/discretisedfield/blob/053de0c2b94c17d00458b4bc31bb3c0edf61af35/discretisedfield/mesh.py#L2149

Edit: we only use `fftfreq` to find `f_min` and `f_max`. It might be useful to add a comment to `mesh.fftn` stating why `fftshift` is not required.

lang-m (commented):
Looks good. I left two small comments. We can merge after these are resolved.

One more general question: do you know why we use `scipy.fft` instead of `numpy.fft`?

discretisedfield/tests/test_field.py
samjrholt: I think it could be us just over-checking or left over from debugging. Would you like me to remove it?

PR #417, opened 2023-08-03
component tests

samjrholt:


discretisedfield/tests/test_field.py
```diff
@@ -843,20 +843,27 @@ def test_mean():
...
+    assert all(getattr(f, i).nvdim == 1 for i in valid_components[:nvdim])
 
-    f = df.Field(valid_mesh, nvdim=2, value=(1, 2))
-    assert all(isinstance(getattr(f, i), df.Field) for i in "xy")
-    assert all(getattr(f, i).nvdim == 1 for i in "xy")
+    # Default
+    f = df.Field(valid_mesh, nvdim=nvdim)
+    if nvdim > 3:
+        valid_components = [f"v{i}" for i in range(nvdim)]
+    elif nvdim > 1:
```
swapneelap, line 857: A small design choice: to me it looks more readable if we have:
```python
if nvdim in [2, 3]:
    ...
elif nvdim > 3:
    ...
else:
    ...
```
But it is a personal choice. You can keep the tests as it is since the logic works :).

swapneelap (commented):
I have a small comment; looks good to me! Please merge it when you like.

discretisedfield/tests/test_field.py
samjrholt: Cheers! I've changed it.

PR #420, opened 2023-08-04
dir tests

samjrholt:
Need to decided what todo with `dir` as I think consistency would be good


swapneelap (approved):
Looks good to me 👍🏽

PR #428, opened 2023-08-04
Review and update `Field` operation documentation

swapneelap:


samjrholt (approved):
@swapneelap  i've made a few extra changes. Unfortunately it wouldn't let me do this in comments as it was too far from the diff. Take a look and let me know or merge if you are happy.

PR #429, opened 2023-08-04
ignore plotting cell

samjrholt:
@swapneelap I missed this before when reviewing but we can make the tests pass by adding a command to the cell. Merge if you are happy.


swapneelap:
@samjrholt I am confused why [this](https://github.com/ubermag/discretisedfield/blob/b89b514d4785df673825b371877e816b95979f53/nbval.cfg#L22-L24) line form `nbval.cfg` is not taking care of the problem? Maybe we can have a look at it separately; I'll merge this for now, thank you for pointing it out.

PR #434, opened 2023-08-04
ufunc tests

samjrholt:


discretisedfield/field.py
```diff
@@ -3790,27 +3790,36 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
             return None
         else:
-            return self.__class__(
-                mesh[0],
-                nvdim=result.shape[-1],
-                value=result,
-                vdims=self.vdims,
-                vdim_mapping=self.vdim_mapping,
-            )
+            print("Hi")
```
swapneelap, line 3810: Maybe remove this 😉

discretisedfield/field.py
```diff
@@ -3790,27 +3790,36 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+            if not np.array_equal(result.shape[:-1], mesh[0].n):
```
swapneelap, line 3811: Can we also do it like this?
```suggestion
            if not np.array_equal(result.shape[:-1], self.mesh.n):
```

discretisedfield/field.py
```diff
@@ -3790,27 +3790,36 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+                raise NotImplementedError()
+            try:
+                return self.__class__(
+                    mesh[0],
```
swapneelap, line 3815: Here as well, can we do this?
```suggestion
                    self.mesh,
```

swapneelap (commented):
Looks good to me. I have left a few small comments.

discretisedfield/field.py
samjrholt, line 3810:
```suggestion
```

PR #448, opened 2023-08-09
Berg-Luescher valid

samjrholt:


discretisedfield/tests/test_tools.py
```diff
@@ -42,8 +42,22 @@ def test_topological_charge(method):
     assert q.nvdim == 1
     assert q.mean() > 0
     for absolute in [True, False]:
+        print(test_filename)
```
swapneelap, line 45: Debugging line to be removed

swapneelap (commented):
Does `topological_charge` and `topological_charge_density` fail silently if method is neither continuous nor Berg-Luescher?

PR #450, opened 2023-08-10
colorbar scaling

samjrholt:
@swapneelap  this should hopefully fix the colorbar issues in the tutorials

### Summary by CodeRabbit

- New Feature: Enhanced the `_add_colorbar` method in `mpl_field.py` to support more customization options. Users can now specify minimum height, width, and padding for the colorbar. The update also allows for the addition and precise placement of multiple colorbars with consistent dimensions. This feature improves the flexibility and aesthetics of data visualization in plots.


lang-m:
This is a significant improvement! Thanks for looking into it.

Two things that I don't currently like (see output of cell 52, cut planes in xy and yz):
- The additional axes are not compatible with tight layout (presumably also not constrained layout), as indicated in the warning
- All colorbar axes appear on the right, not next to the axes object they belong to. So relating colorbars to axes is difficult and potentially impossible if the same colormap is used multiple times.

samjrholt:
Thanks Martin. Yes there are still a few issues that I need to fix. Including the fact that if you do a subplot of any shape everything will end up on the right!!

discretisedfield/plotting/mpl_field.py
```diff
@@ -872,9 +872,87 @@ def __dir__(self):
...
+
+        if min_width_norm > 0.05:
+            width_h = Size.Fixed(min_width_inches)
+        else:
+            width_h = Size.AxesX(ax, aspect=0.05)
+
+        if min_height_norm > 1:
+            v_aspect = min_height_norm
+        else:
+            v_aspect = 1
```
lang-m, line 917: I would find some additional comments explaining the ideas behind the individual parts useful.

discretisedfield/plotting/mpl_field.py
```diff
@@ -872,9 +872,87 @@ def __dir__(self):
...
+        # divider.set_horizontal(h)
+        divider.set_vertical(v)
+
+        ax.set_axes_locator(divider.new_locator(nx=0, ny=0))
+
+        for i, cb in enumerate(existing_colorbars):
+            cb.set_axes_locator(divider.new_locator(nx=2 * (i + 1), ny=0))
+        cax.set_axes_locator(
+            divider.new_locator(nx=2 * (len(existing_colorbars) + 1), ny=0)
+        )
```
lang-m, line 953: What is done in this part of the code?

lang-m (commented):
In terms of functionality and usability this is a huge step forward. There are still a few problems, see below. Depending on how much time we have, I would be open to merge this before fully addressing the problem of partially overlapping colorbars and instead open an issue for this.

Some additional comments explaining the individual parts of the code would be useful.

There are still some (less severe) problems with overlapping colorbars, e.g. in this example: 
image
[We do not see the problem in the notebook because we increase `figsize` in the notebook.]

discretisedfield/plotting/mpl_field.py
samjrholt: So when we add new axes, we go through and associate the each old axes with a location as a index and finally the new colour bar

samjrholt:
> In terms of functionality and usability this is a huge step forward. There are still a few problems, see below. Depending on how much time we have, I would be open to merge this before fully addressing the problem of partially overlapping colorbars and instead open an issue for this.
> 
> Some additional comments explaining the individual parts of the code would be useful.
> 
> There are still some (less severe) problems with overlapping colorbars, e.g. in this example: image [We do not see the problem in the notebook because we increase `figsize` in the notebook.]

I will open a new issue for this

lang-m:
Overlapping colorbars tracked in #450

discretisedfield/plotting/mpl_field.py
```diff
@@ -872,9 +872,123 @@ def __dir__(self):
...
+        # The main plot's axes will be placed at the starting (0,0) position
+        # on the grid defined by the divider.
+        ax.set_axes_locator(divider.new_locator(nx=0, ny=0))
+
+        # Loop through any existing colorbars associated with the main axes.
+        # Adjust their positions to accommodate the new colorbar.
+        # Each colorbar is positioned two steps away on the x-axis
+        # (e.g., main plot at 0, first colorbar at 2, second at 4, and so on).
+        for i, cb in enumerate(existing_colorbars):
+            cb.set_axes_locator(divider.new_locator(nx=2 * (i + 1), ny=0))
```
lang-m, line 980:
```suggestion
        for i, cb in enumerate(existing_colorbars, start=1):
            cb.set_axes_locator(divider.new_locator(nx=2 * i, ny=0))
```

discretisedfield/plotting/mpl_field.py
```diff
@@ -872,9 +872,123 @@ def __dir__(self):
...
+
+        # Set the location of the main plot's axes to its original position.
+        # The main plot's axes will be placed at the starting (0,0) position
+        # on the grid defined by the divider.
+        ax.set_axes_locator(divider.new_locator(nx=0, ny=0))
+
+        # Loop through any existing colorbars associated with the main axes.
+        # Adjust their positions to accommodate the new colorbar.
+        # Each colorbar is positioned two steps away on the x-axis
+        # (e.g., main plot at 0, first colorbar at 2, second at 4, and so on).
```
lang-m, line 978: What is the reason for the even steps?

samjrholt: This is as the pads are the odd numbers

lang-m (approved):
I think we can merge this now. Or do you want to add anything @samjrholt ?

PR #451, opened 2023-08-10
plot on valid

samjrholt:
@lang-m since the change the ipynb tests are failing. This looks like due to a different output (could be my laptop). Should we add these to `nbval` or do you have another idea?

### DIFF
--- a/discretisedfield/field.py
+++ b/discretisedfield/field.py
@@ -3536,7 +3536,12 @@ class Field(_FieldIO):
         :DynamicMap...
 
         """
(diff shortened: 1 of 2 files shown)


lang-m (commented):
Best to discuss in person: I think we should keep `roi` and `valid` separate and only return `valid` data in `_hv_data_selection`. The two have slightly different functionality, e.g. we could look at the top left quadrant (-> `roi`) of a sphere (-> `valid`).

lang-m:
Closing in favour of #475

PR #461, opened 2023-08-21
Test mpl

samjrholt:


samjrholt:
@lang-m @coderabbit when plotting if we try to plot the incorrect number of `ndim` then we get a `RuntimeError` but if we try to plot the incorrect number of `nvdims` we get a `ValueError`. What are you opinions on standardising this and which should it be?

samjrholt:
> @lang-m when plotting if we try to plot the incorrect number of `ndim` then we get a `RuntimeError` but if we try to plot the incorrect number of `nvdims` we get a `ValueError`. What are you opinions on standardising this and which should it be?

@coderabbitai

discretisedfield/tests/test_field.py
```diff
@@ -3151,14 +3151,15 @@ def test_mpl(test_field):
...
-    field = df.Field(valid_mesh, nvdim=1)
+@pytest.mark.parametrize("nvdim", [1, 2, 3, 4])
+def test_mpl_dimension(valid_mesh, nvdim):
+    field = df.Field(valid_mesh, nvdim=nvdim)
 
-    if valid_mesh.region.ndim != 2:
+    if valid_mesh.region.ndim != 2 or nvdim not in [1, 2, 3]:
         with pytest.raises(RuntimeError):
-            field.mpl.scalar()
+            field.mpl()
```
lang-m, line 3160: For a wrong number of `vdims` we get a different error:
```python
>>> (field.sel(z=10e-9) << field.sel(z=10e-9)).mpl()
...
ValueError: At least one element of vdims=[None, None] must be not None.
```

Side note, not to be added to the test:
Plotting would still work for higher-dimensional fields when we specify `vdims`:
```python
(field.sel(z=10e-9) << field.sel(z=10e-9)).mpl(vector_kw={'vdims': ['v2', 'v5']})
```
Only a subset of 3 vdims is then shown in the plot.

lang-m (changes requested):

discretisedfield/tests/test_field.py
```diff
@@ -3160,14 +3160,18 @@ def test_mpl(test_field):
...
+def test_mpl_dimension(valid_mesh, nvdim):
+    field = df.Field(valid_mesh, nvdim=nvdim)
 
     if valid_mesh.region.ndim != 2:
         with pytest.raises(RuntimeError):
-            field.mpl.scalar()
+            field.mpl()
+    elif nvdim not in [1, 2]:
+        with pytest.raises(ValueError):
+            field.mpl()
```
lang-m, line 3172: Unfortunately, this is much more complicated: we can plot higher-dimensional vectors if `field.vdim_mapping` is defined. This `dict` maps `vdims` to `dims`. Internally, we swap keys and values to find `vdims` that point along the two `dims` of our field.mesh (the axes of our plot). Additionally, we use the first (or maybe one random?) additional component for colour - with the idea of the implementation of just having 3 vdims, and using colour for the single out-of-plane component (the logic is probably from before our changes). We should consider raising some sort of error (maybe `RunTime`) when nvdim > 3. Thoughts?

samjrholt: `mpl()` is mainly meant to be for convenience. However, when `nvdims>3` there is no unique choice for the scalar coloring. Hence, I agree with changing the implementation to raise a `RuntimeError`.

lang-m (approved):

lang-m (approved):
Approved once all tests pass.

samjrholt:
> Approved once all tests pass.

I'm really sorry I didn't spot that it failed!! I've made an edit, let me know your thoughts as I'm less sure if this is the solution we want

discretisedfield/tests/test_field.py
```diff
@@ -3160,14 +3160,19 @@ def test_mpl(test_field):
...
+    if valid_mesh.region.ndim != 2 or nvdim > 3:
         with pytest.raises(RuntimeError):
-            field.mpl.scalar()
+            field.mpl()
     else:
-        field.mpl.scalar()
+        if nvdim == 3:
+            field.vdim_mapping = dict(
+                zip(field.vdims, [*valid_mesh.region.dims, None, None])
+            )
```
lang-m, line 3174: I think this is in principle good.

We should only need one additional `None` as `dims==2`.
```suggestion
            field.vdim_mapping = dict(
                zip(field.vdims, [*valid_mesh.region.dims, None])
            )
```

PR #462, opened 2023-08-22
FFT tests

samjrholt:
### Summary by CodeRabbit

- Bug Fix: Corrected the calculation of `dfreq` in `fftn` and `ifftn` methods to ensure it's always positive, enhancing the reliability of Fourier Transform operations.
- Test: Enhanced the `test_fft` function for better coverage and accuracy. Added two new test functions, `test_rfft_no_shift_last_dim` and `test_1d_fft`, to validate rFFT operation without shifting the last dimension and 1D FFT operation respectively, improving the robustness of the testing suite.

### DIFF
--- a/discretisedfield/mesh.py
+++ b/discretisedfield/mesh.py
@@ -2248,7 +2248,7 @@ class Mesh(_MeshIO):
                 # Shift the region boundaries to get the correct coordinates of
                 # mesh cells.
                 # This effectively does the same as using fftshift
-                dfreq = (freqs[1] - freqs[0]) / 2
+                dfreq = abs(freqs[1] - freqs[0]) / 2
                 p1.append(min(freqs) - dfreq)
                 p2.append(max(freqs) + dfreq)
                 n.append(len(freqs))
@@ -2392,7 +2392,7 @@ class Mesh(_MeshIO):
                 freqs = spfft.fftfreq(shape[i], self.cell[i])
                 # Shift the region boundaries to get the correct coordinates of
                 # mesh cells.
-                dfreq = (freqs[1] - freqs[0]) / 2
+                dfreq = abs(freqs[1] - freqs[0]) / 2
                 p1.append(min(freqs) - dfreq)
                 p2.append(max(freqs) + dfreq)
                 n.append(len(freqs))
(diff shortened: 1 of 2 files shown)


discretisedfield/mesh.py
```diff
@@ -2248,7 +2248,7 @@ def fftn(self, rfft=False):
                 # Shift the region boundaries to get the correct coordinates of
                 # mesh cells.
                 # This effectively does the same as using fftshift
-                dfreq = (freqs[1] - freqs[0]) / 2
+                dfreq = abs(freqs[1] - freqs[0]) / 2
```
lang-m, line 2251: This change seems reasonable. Why did we not have (need) it before?

lang-m (approved):
These changes look good and can be merged.

However, we only test that fftn -> ifftn results in the original field. We should maybe add one additional tests for a very simple case (e.g. scalar 1d field with a few points), where we check the actual fft values against manually computed ones. Thoughts?

discretisedfield/mesh.py
samjrholt: There was an assumption that `freqs` is always increasing however if we do an fft on a small mesh I found this wasn't the case.

discretisedfield/tests/test_field.py
```diff
@@ -2933,6 +2933,22 @@ def test_rfft_no_shift_last_dim():
...
+    field_ft = f.fftn()
+    expected_array = np.array(
+        [
+            -5.0 - 1.62459848j,
+            -5.0 - 6.8819096j,
+            25.0 - 0.0j,
+            -5.0 + 6.8819096j,
+            -5.0 + 1.62459848j,
+        ]
+    )
```
lang-m, line 2948: How did you compute the reference values?

samjrholt: I used `numpy.fft` as we use scipy in the code

lang-m: Should we then maybe just add this to the test instead of the hard-coded values? We are anyways not interested in precisely testing the values and instead mainly want to ensure that we compute something reasonable (i.e. an FFT at all).

PR #468, opened 2023-08-30
create valid docs

samjrholt:


lang-m (commented):
This overall looks good to me. 

Should we:
- be more specific during which operations valid is used/ignored
- add an example for a different operation, e.g. adding two fields with different `valid` (in which case, I assume, we use `valid` from the left field?)

lang-m:
@samjrholt How do we want to proceed with this?

lang-m:
The changes look good. Final question: should we mention the word `intersection` in

> When you perform an operation on these fields, the resulting `valid` attribute is determined by the logical AND operation on the `valid` attribute of the initial fields.

Please do as you think is best and merge afterwards.

PR #478, opened 2023-09-20
Mpl vector

samjrholt:


discretisedfield/tests/test_field.py
```diff
@@ -3085,6 +3085,26 @@ def test_mpl_vector(test_field):
...
+def test_mpl_dimension_vector(valid_mesh, nvdim):
+    field = df.Field(valid_mesh, nvdim=nvdim)
+
+    if valid_mesh.region.ndim != 2:
+        with pytest.raises(RuntimeError):
+            field.mpl.vector()
+    else:
+        if nvdim != 2:
+            with pytest.raises(ValueError):
+                field.mpl.vector()
```
lang-m, line 3098: `mpl.vector` also works for `nvdim=3`. We use colour to show the out-of-plane component. Additionally, any higher `nvdim` is also allowed if `vdim_mapping` is defined (the latter is also required for `nvdim in [1, 2, 3]`). For high-dimensional fields we do not do automatic colouring and only show the two components in the plot plane obtained from `vdim_mapping` (or specified explicitly using the `vdims` argument of `mpl.vector`).

discretisedfield/tests/test_field.py
```diff
@@ -3085,6 +3085,27 @@ def test_mpl_vector(test_field):
...
+
+    if valid_mesh.region.ndim != 2:
+        with pytest.raises(RuntimeError):
+            field.mpl.vector()
+    else:
+        if nvdim not in [2, 3]:
+            with pytest.raises(ValueError):
+                field.mpl.vector()
+        else:
+            field.vdim_mapping = dict(zip(field.vdims, [*valid_mesh.region.dims, None]))
```
lang-m, line 3100: Setting a vdim mapping and then using `mpl.vector` should in principle also work for `nvdim in [1, 4]`. It is however probably not overly important to add the tests for this as it is rarely needed.

We should however test, that plotting works without setting `vdim_mapping` manually if `ndim == nvdim` for the cases 2, 3 and 4.

lang-m (changes requested):

discretisedfield/tests/test_field.py
samjrholt: Everything that does not have `ndim==2` will give a `RuntimeError` by design.  `vdim_mapping` is required for `mpl.vector` but is not generated automatically for all `nvdim` as `ndim==2`, hence the manual setting. 

However, I agree this should work for more cases i.e. `nvdim>1` but I don't think it should work for `nvdim==1`, hence have made this edit.

samjrholt: `test_mpl_lightess` is also testing the lightness function starting with a 3D field and selecting planes.

lang-m: `nvdim==1` of course does not work for vector, you are right. The case `nvdim==2` should work without setting `vdim_mapping`, right? Maybe we can split this one off?

samjrholt: Yeah you're right that this should work automatically for `nvdim==2`. I'm slightly confused at why we would want to split it off though. Granted we are resetting something that is already set, but this means we are trying to make the tests as orthogonal as possible.  This test will pass even if there is an issue setting `vdim_mapping` automatically and only the `test_vdims_vdim_mapping` test would fail.

lang-m: Okay, let us merge as is.

lang-m (approved):

PR #490, opened 2023-09-22
test(Field): HoloViews plotting methods for arbitrary ndim and nvdim

lang-m:


lang-m:
The failing `flake8` (in the pre-commits) passes locally and the error message seems wrong. @samjrholt Could you please confirm whether the error makes any sense or not.

samjrholt (approved):

PR #494, opened 2023-09-28
Attempt: fix holoviews - xarray bug

lang-m:
- Test with Python 3.8 and 3.10
- Fix holoviews vectorplot bug; this was actually a problem with xarray and numpy ufuncs when computing the inplane-angle using `np.arctan2`
- Hide FutureWarning and UserWarning from HoloViews

### DIFF
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -4,7 +4,6 @@ channels:
   - conda-forge
 
 dependencies:
-  - python==3.8
   - pip
   - oommf
   - pip:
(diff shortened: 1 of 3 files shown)


lang-m:
@samjrholt I have disabled pre-commits for discretisedfield. I hope that maybe when we re-activate it later (after merging) the wrong error might disappear.

samjrholt (approved):
This looks good to me. Thank you for linking to the appropriate place in holoviews.

PR #506, opened 2023-11-09
ignore hv cell

samjrholt:
One cell in the field holoviews plotting started failing with Python 3.10 (no obvious error, probably some sort of change in the format). As a workaround we ignore that cell.


samjrholt:
@lang-m this has fixed the issue. Please merge and update the latest tag if you are happy.

PR #508, opened 2023-11-17
Rename `mesh.points` -> `mesh.cells`

swapneelap:


swapneelap:
Should we add a deprecation warning for the 'points' attribute to guide users towards the new 'cells' attribute by re-adding the points property that returns the cells attribute and raises a DeprecationWarning as the bot says? @samjrholt @lang-m

samjrholt:
> Should we add a deprecation warning for the 'points' attribute to guide users towards the new 'cells' attribute by re-adding the points property that returns the cells attribute and raises a DeprecationWarning as the bot says? @samjrholt @lang-m

I would be in favour of using `DeprecationWarning` in general as it makes it easier for the users to transition.

lang-m:
I would re-add `points` to make the transition easier. What we could consider however is raising an `AttributeError` with a meaningful message that points to `cells` (instead of just the warning). That is what we did for most breaking changes in the previous release. This is of course quite harsh but forces quick adoption of the new interface. Thoughts (also @fangohr)?

PR #510, opened 2023-11-22
Convert `as_array` to a method

lang-m:
To prepare cell and vertex fields we convert `_as_array` to a method so that derived classes can override it


swapneelap (approved):
Looks good to me!

PR #511, opened 2023-11-22
Add `VertexField` to support field data at vertices

swapneelap:


swapneelap:
@lang-m the hvplot should be working with the `VertexField` now. Play with it if you find some time and let me know if something looks odd.

PS: Thank you for setting up `hv` in a way that small modification to the `to_xarray` fixed the whole plotting for `VertexField` 😃

discretisedfield/__init__.py
```diff
@@ -6,14 +6,16 @@
 import matplotlib.pyplot as plt
 import pytest
 
-from . import tools
+from . import cell_field, tools, vertex_field
+from .cell_field import CellField
```
samjrholt, line 10:
```suggestion
```

discretisedfield/__init__.py
```diff
@@ -6,14 +6,16 @@
...
 from .field import Field
 from .field_rotator import FieldRotator
 from .interact import interact
 from .line import Line
 from .mesh import Mesh
 from .operators import integrate
 from .region import Region
+from .vertex_field import VertexField
```
samjrholt, line 18:
```suggestion
```

discretisedfield/field.py
```diff
@@ -157,12 +156,36 @@ class Field(_FieldIO):
         "write": "to_file",  # method is in io.__init__
     }
 
+    def __new__(
+        cls,
+        mesh,
+        nvdim=None,
+        value=0.0,
+        norm=None,
+        data_location="cell",
```
samjrholt, line 165: Are we sure about the name? (Just checking :+1:  )

samjrholt (commented):
As discussed, we should use `abstractmethod` rather than overwriting the base class

- [ ] mpl
- [x] to_xarray
- [x] from_xarray
- [ ] _hv_key_dims

PR #513, opened 2023-11-24
Fix: more general ovf reading

lang-m:
Currently, we take capitalisation of the data line in the ovf header into account, however that does not seem to be well-defined in the standard. (We use the same format that OOMMF writes but the documentation on the OOMMF website uses a different standard.)

To make it more flexible, we can ignore the case.

Here are the specifications: https://math.nist.gov/oommf/doc/userguide21a0/userguide/Data_block.html

Closes ubermag/help#266


lang-m:
@samjrholt and @swapneelap I would like to merge this as soon as possible and release a small bug-fix
I will add the file provided in the issue as additional test file if I get the permission.

samjrholt (approved):

swapneelap (approved):
Looks good to me 👍🏽

PR #518, opened 2024-03-19
Test deprecation warning

kzqureshi:


discretisedfield/line.py
```diff
@@ -170,7 +170,7 @@ def point_columns(self):
 
     @point_columns.setter
     def point_columns(self, val):
-        if len(val) != 3:
+        if len(val) != len(self._point_columns):
```
lang-m, line 173: This is part of #524; please remove  from this MR.

pytest.ini
```diff
@@ -0,0 +1,9 @@
+[pytest]
+python_files = test_*.py
+python_functions = test_*
+filterwarnings =
+    ignore::DeprecationWarning:traitlets.*:
+    #ignore:Automatic coloring is only supported for 3d fields.*:UserWarning
+    ignore:The truth value of an empty array is ambiguous.*:DeprecationWarning
+    ignore:invalid value encountered in divide:RuntimeWarning
+    ignore:`ipykernel.pylab.backend_inline` is deprecated.*:DeprecationWarning
```
lang-m, line 9: - We don't want this file, instead everything should go into `pyproject.toml`
- I am not sure if all these exceptions should really be defined. We should only ignore warnings in third-party libraries that are not caused by us.

discretisedfield/tools/tools.py
```diff
@@ -273,9 +273,17 @@ def topological_charge(field, /, method="continuous", absolute=False):
 
     q = topological_charge_density(field, method=method)
     if absolute:
-        return float(abs(q).integrate())
+        result = abs(q).integrate()
+        if isinstance(result, np.ndarray) and result.size == 1:
+            result = result.item()
+        assert np.isscalar(result), "Expected a scalar result from integration"
+        return float(result)
```
lang-m, line 280: What are these conversions required for?

lang-m (changes requested):

PR #519, opened 2024-03-23
Ruff additional rules

lang-m:
Most changes have been made by ruff fix and ruff formatting. I added ignores for lines that I would not want to change for some reason.

If everything in here looks reasonable to you, I would merge this, copy the changes to the devtools template and then update all other repos based on this.


swapneelap (approved):
Looks good to me! 👍🏽

PR #524, opened 2024-04-26
added a check for n dimensional field in point_columns

kzqureshi:


lang-m (changes requested):
We cannot rely on `self._point_columns` for this test. Instead we should use the shape of points for this check. Further details are best discussed in person.

PR #543, opened 2026-06-05
solved test issues with k3d in test_field.py, test_mesh.py, test_region

AdamsMP92:
````md
## Summary

This PR updates the `k3d`-related tests to avoid calling `k3d.Plot.display()` during pytest runs.

Instead of relying on the plotting methods to create and display a `k3d` plot internally, the tests now create an explicit `k3d.plot()` object and pass it via the `plot` argument.

## Motivation

With `k3d==2.17.0`, calling

```python
k3d.plot().display()
````

fails in a non-IPython pytest session with

```text
NameError: name 'display' is not defined
```

because `k3d.plot.plot_display` calls `display(...)` without importing it from `IPython.display`.

The purpose of these tests is to exercise the `discretisedfield` k3d plotting code paths, not to test whether the interactive Jupyter display machinery is available in a terminal pytest environment.

## Changes

* Pass explicit `k3d.plot()` objects to k3d plotting calls in tests.
* Avoid direct calls to `plot.display()` in tests.
* Keep the existing k3d plotting paths covered for `Field`, `Mesh`, and subregion plotting.

## Test result

The local test suite now passes:

```text
3636 passed, 2 skipped, 5 xfailed in 40.11s
```

### DIFF
--- a/discretisedfield/tests/test_field.py
+++ b/discretisedfield/tests/test_field.py
@@ -3737,22 +3737,22 @@ def test_k3d(valid_mesh):
     f = df.Field(valid_mesh, nvdim=3, value=(1, 1, 1))
     if f.mesh.region.ndim != 3:
         with pytest.raises(RuntimeError):
-            #f.k3d.vector()
+            # f.k3d.vector()
             plot = k3d.plot()
             f.k3d.vector(plot=plot)
     else:
-        #f.k3d.vector()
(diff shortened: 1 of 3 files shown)


lang-m:
The reason for the failure is a bug in k3d 2.17. #544 excludes that problematic version.

